PR #6, opened 2018-08-11
A unified contributing experience.

dvc94ch:


japaric:
@dvc94ch tests are failing. This won't build on stable or beta because the crate uses unstable features.

japaric:
> tests are failing.

See https://travis-ci.org/rust-embedded/riscv/builds/414919869?utm_source=github_status&utm_medium=notification

danc86:
I've been (very slowly) picking at the various unstable bits... Do you think we need to get them all fixed before we can merge this?

Personally, I would be fine to merge this and just leave the stable and beta tests failing for now. Worst case we can just ignore the failures. But it will also help me to know when I've finally got it fixed :-)

dvc94ch:
@danc86 I disabled it for now. What are your thoughts on licenses? I detest having to have to large legal texts in a software repo. But if we are going to relicense to Apache+MIT it's now or never =P

danc86:
> What are your thoughts on licenses?

Personally I always pick GPLv2+ for my code, because I'm a copyleft hippie, but I guess Apache+MIT is the standard throughout the Rust ecosystem. Probably worth sticking with that.

I'll leave it up to you. Take this as my approval to relicense my contributions to all the riscv-rust repos under any license of your choosing.

PR #9, opened 2018-08-18
use `NonZeroUsize` where appropriate

strake:


danc86:
The failed merge build was just totally empty... I guess some kind of flake within Travis CI?

Let's see if we can make bors try it again...

bors retry

PR #13, opened 2018-12-18
Replace no-op with unimplemented!()

Disasm:


Disasm:
Thanks. Should we trigger bors somehow?

PR #19, opened 2019-02-06
Add team e-mail to authors

Disasm:


dvc94ch:
Can you please remove me from the authors? I think that is more appropriate for a team effort.

Disasm:
@dvc94ch As you wish. However, I was guided by [discussion in IRC](https://github.com/rust-embedded/wg/blob/f85ce1c00b51f03def00813c13baeb235136e6a6/minutes/2019-01-08.irc.log#L131-L140) and [common practice](https://github.com/rust-embedded/cortex-m-rt/commit/6cdfa65520b96691b50098249e72d82fffabd112#diff-80398c5faae3c069e4e6aa2ed11b28c0), so If you still want to be removed from the authors, let me know. Also, should I do the same [for riscv-rt crate](https://github.com/rust-embedded/riscv-rt/pull/21)?

Disasm:
BTW, thanks for adding to the riscv-rust org. I am just planning to update crates that are hosted there. Can I just push my changes instead of creating PRs? (I will check build with CI scripts before that)

dvc94ch:
The git history is the authority on who did what. The authors list may be out of date, incomplete or missleading. I think we should just have the team email.

If you think it's an uncontroversial change and you tested it I'm fine with you pushing directly.

PR #20, opened 2019-02-19
move doc comments inside macro invocations

euclio:
[rust-lang/rust#57882](https://github.com/rust-lang/rust/pull/57882) is modifying the `unused_doc_comments` lint to fire on mistakenly documented macro expansions. Note that these doc comments are not currently used, since they are eliminated when the macro is expanded. A crater run detected that this crate will break due to this change, likely because of the use of `deny(unused_doc_comments)` or `deny(warnings)`.

While this kind of breakage is allowed under Rust's stability guarantees, I am opening PRs to affected crates to reduce the impact.

This PR protects your crate from future breakage by moving the offending doc comments inside the macro invocations.

### DIFF
--- a/src/register/macros.rs
+++ b/src/register/macros.rs
@@ -183,7 +183,8 @@ macro_rules! clear {
 }
 
 macro_rules! set_csr {
-    ($set_field:ident, $e:expr) => {
+    ($(#[$attr:meta])*, $set_field:ident, $e:expr) => {
+        $(#[$attr])*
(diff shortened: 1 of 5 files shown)


Disasm:
Thanks for the info! I think it's better to use these doc comments like how it's done here: https://github.com/rust-embedded/riscv/blob/master/src/asm.rs

euclio:
Fixed, thanks.

PR #30, opened 2019-06-08
add riscv32i target

sbourdeauducq:
This adds support for the simplest RISC-V processor possible, which is useful e.g. for experimenting with  FPGA softcore implementations.


Disasm:
Hmm. I do not see `riscv32i-unknown-none-elf` target in Rust. How are you going to use this?

sbourdeauducq:
Patch the Rust compiler as well, e.g. https://git.m-labs.hk/M-Labs/HeavyX/commit/033659344f1bdfcc893fee13156bb224872bb4a4

Disasm:
Could you upstream this first? It should be an easy task.

Disasm:
Also double-check that you are targeting softcores with Privileged Architecture Version 1.10 support. These crates (`riscv` and `riscv-rt`) are pretty useless for cores without it (e.g. `picorv32`).

Disasm:
I [added riscv32i-unknown-none-elf to Rust](https://github.com/rust-lang/rust/pull/62784), so I can merge this in a day if you still need it. However, this PR needs to be rebased first.

sbourdeauducq:
Thanks!
Will rebase in a moment.

sbourdeauducq:
Couldn't test since https://github.com/rust-embedded/riscv-rt/pull/31 broke https://github.com/rust-embedded/riscv-rt/pull/34

Disasm:
Yes, we need to figure out what to do with that mul instruction on riscv32i

laanwj:
> Yes, we need to figure out what to do with that mul instruction on riscv32i

As it'd be only a small iteration up to # of cores, won't lose much by replacing it with a loop, maybe conditionally on non-M.

Disasm:
> As it'd be only a small iteration up to # of harts, won't lose much by replacing it with a loop, maybe conditionally on non-M.

Sure, the change is not difficult, but it needs to be done for corresponding `riscv-rt` PR.

PR #31, opened 2019-06-25
Add more CSRs

Disasm:


laanwj:
> or does the one-function-per-section cause only the used functions to be linked in in the final binary?

I've checked and this is indeed the case ! Only what is actually used makes it into the binary.

bors r+

Disasm:
Yep, this is quite common pattern for saving space. In two Rust releases we will have even more beautiful solution with ThinLTO.

PR #32, opened 2019-08-27
add user trap setup and handling registers

almindor:
Adds CSR handlers for user trap setup and handling registers.

Also unifies common shared types.

NOTE: untested, will test on real device ASAP

### DIFF
--- a/src/register/mod.rs
+++ b/src/register/mod.rs
@@ -13,11 +13,18 @@
 #[macro_use]
 mod macros;
 
-// TODO: User Trap Setup
+// User Trap Setup
+// TODO: sedeleg, sideleg
+pub mod ustatus;
+pub mod uie;
+pub mod utvec;
 
-
-// TODO: User Trap Handling
-
+// User Trap Handling
+pub mod uscratch;
+pub mod uepc;
+pub mod ucause;
+pub mod utval;
+pub mod uip;
 
 // User Floating-Point CSRs
 // TODO: frm, fflags
--- a/src/register/mod.rs
+++ b/src/register/mod.rs
@@ -14,7 +14,6 @@
 mod macros;
 
 // User Trap Setup
-// TODO: sedeleg, sideleg
 pub mod ustatus;
 pub mod uie;
 pub mod utvec;
--- a/src/register/scause.rs
+++ b/src/register/scause.rs
@@ -3,80 +3,12 @@
 use bit_field::BitField;
 use core::mem::size_of;
 
+pub use crate::register::mcause::{Interrupt, Exception, Trap};
+
 /// scause register
 #[derive(Clone, Copy)]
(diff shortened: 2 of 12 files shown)


src/register/scause.rs
```diff
@@ -3,82 +3,14 @@
 use bit_field::BitField;
 use core::mem::size_of;
 
+pub use crate::register::mcause::{Interrupt, Exception, Trap};
```
Disasm: `Interrupt` and `Exception` should be different for `mcause` and `scause`

src/register/ucause.rs
```diff
@@ -0,0 +1,60 @@
+//! ucause register
+
+pub use crate::register::mcause::{Interrupt, Exception, Trap};
```
Disasm: Same here. Is there any information about `scause` register in specs?

src/register/ustatus.rs
```diff
@@ -0,0 +1,76 @@
...
+            0b10 => XS::NoneDirtySomeClean,
+            0b11 => XS::SomeDirty,
+            _ => unreachable!(),
+        }
+    }
+
+    /// Whether either the FS field or XS field
+    /// signals the presence of some dirty state
+    #[inline]
+    pub fn sd(&self) -> bool {
```
Disasm: Are `FS`, `XS` and `SD` present in the "restricted views" of `mstatus`? Can't find anything about this.

src/register/mod.rs
```diff
@@ -13,11 +13,18 @@
 #[macro_use]
 mod macros;
 
-// TODO: User Trap Setup
-
-
-// TODO: User Trap Handling
-
+// User Trap Setup
+// TODO: sedeleg, sideleg
```
Disasm: There is no `sedeleg` and `sideleg` counterparts for user mode

src/register/scause.rs
almindor: I see the reserved fields pose a problem in the future, reverted

src/register/ustatus.rs
almindor: removed since I don't think you should get that info in user mode, not sure tho...

src/register/ucause.rs
almindor: Asking around now, seems like it's blatantly missing.

almindor: Made ucause only readable as bits since the `N` extension doesn't seem to be finalized.

laanwj:
LGTM too

> NOTE: untested, will test on real device ASAP

Just curious: is there any known real device that does user traps?

almindor:
Good question, the hifive1 devices support user mode but I didn't try/check if they actually have user traps. The docs just refer to the RISCV manuals AFAICS.

PR #33, opened 2019-09-05
Do not use bare-metal v0.2.5

Disasm:


Disasm:
See also: https://github.com/rust-embedded/riscv/pull/32

PR #34, opened 2019-10-09
mip, satp: Allow writing to these privileged registers

archshift:
Per the RISC-V privileged ISA, these registers should be writable.

### DIFF
--- a/src/register/mip.rs
+++ b/src/register/mip.rs
@@ -71,3 +71,30 @@ impl Mip {
 }
 
 read_csr_as!(Mip, 0x344, __read_mip);
+set!(0x344, __set_mip);
+clear!(0x344, __clear_mip);
+
+set_clear_csr!(
+    /// User Software Interrupt Pending
+    , set_usoft, clear_usoft, 1 << 0);
+set_clear_csr!(
+    /// Supervisor Software Interrupt Pending
+    , set_ssoft, clear_ssoft, 1 << 1);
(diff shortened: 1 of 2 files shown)


src/register/mip.rs
```diff
@@ -71,3 +71,30 @@ impl Mip {
...
+    , set_stimer, clear_stimer, 1 << 5);
+set_clear_csr!(
+    /// Machine Timer Interrupt Pending
+    , set_mtimer, clear_mtimer, 1 << 7);
+set_clear_csr!(
+    /// User External Interrupt Pending
+    , set_uext, clear_uext, 1 << 8);
+set_clear_csr!(
+    /// Supervisor External Interrupt Pending
+    , set_sext, clear_sext, 1 << 9);
```
laanwj: there's also a bit 11
the spec calls it MEIP
according to this naming it would be mext/"Machine External Interrupt Pending"

archshift: The spec says that MEIP isn’t writeable, which is why I didn’t include it here

almindor: Correct: ```The MEIP field in mip is a read-only bit that indicates a machine-mode external interrupt is
pending. MEIP is set and cleared by a platform-specific interrupt controller. The MEIE field in
mie enables machine external interrupts when set.```

almindor:
@laanwj I'll let you bors this in case there was something else apart from the MEIP field.

PR #35, opened 2020-02-02
Add mideleg register support

archshift:


archshift:
Hi @Disasm, sorry for the delay! I've gotten rid of the machine-mode bits as you're right, they really don't make sense. I've also added in a structure for reads.

PR #37, opened 2020-03-11
Add support for custom vexriscv instructions

xobs:
This adds a new namespace under "registers" to support the custom interrupt controller on vexriscv.


src/register/vexriscv/mod.rs
```diff
@@ -0,0 +1,10 @@
+pub mod dci;
+pub use self::dci::*;
```
Disasm: `read_csr_as_usize` macro will declare a `read` function in each module. Is it useful to *-reexport all the module contents?

src/register/mod.rs
```diff
@@ -108,3 +108,7 @@ pub use self::mhpmeventx::*;
 
 
 // TODO: Debug Mode Registers
+
+
+// Vexriscv custom instructions
```
Disasm:
```suggestion
// Vexriscv custom CSRs
```

src/register/vexriscv/mod.rs
xobs: No, I misunderstood which modules get re-exported.  It looks like only certain register groups -- namely those that end in `-x` -- get exported.  I'll remove the re-export.

src/register/mod.rs
xobs: Good call

xobs:
1. I've added some module docs.  Please let me know if that's what you're looking for, or if you'd prefer something else.

2. You're correct: https://github.com/SpinalHDL/VexRiscv/blob/master/src/main/scala/vexriscv/plugin/ExternalInterruptArrayPlugin.scala#L25 and https://github.com/SpinalHDL/VexRiscv/blob/95237b23ea2d658cb3e0aa77680ca2851ef5d882/src/main/scala/vexriscv/plugin/DBusCachedPlugin.scala#L358 -- I'll update those to be readonly.

PR #42, opened 2020-05-15
Only checking for necessary extensions when linking

ilya-epifanov:
`M`, `A`, `F` and `D` (as part of `G`) extensions don't matter at all, so generating binaries for them is not necessary, really. Also, if another rust RISC-V target were to come up, this crate would not support it.

In this PR I'm restricting a list of extensions to consider when generating static archives and linking to `I` and `C`.


ilya-epifanov:
The build seems to be broken in `master`. None of the build errors were caused by changes in this PR

PR #47, opened 2020-05-27
Replace asm! with llvm_asm!

Disasm:
`asm!` was replaced with a different implementation on nightly, the old one is renamed into `llvm_asm!`


almindor:
r+

PR #50, opened 2020-06-20
Update MSRV, release v0.6.0

Disasm:

PR #51, opened 2020-06-20
Fix {S,U}tvec::trap_mode() functions to match Mtvec::trap_mode()

Disasm:
See also: https://github.com/rust-embedded/riscv/pull/50#issuecomment-647012575


Disasm:
It's better to release this now, than to make yet another breaking change later.

Disasm:
Published!

PR #52, opened 2020-06-25
Add CSR register `medeleg`; small doc fix

luojia65:
mideleg is at 0x303; medeleg is at 0x302.

### DIFF
--- a/src/register/medeleg.rs
+++ b/src/register/medeleg.rs
@@ -1,0 +1,148 @@
+//! medeleg register
+
+use bit_field::BitField;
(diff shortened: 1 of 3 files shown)


Disasm:
Good catch, thank you! Could you change the doc comments for `medeleg` so that they are written in CamelCase too?

luojia65:
@Disasm Thank you! I've updated documents for `medeleg` to uppercase words.

Disasm:
Seems like rustfmt test failed.

PR #57, opened 2020-10-15
Add mcounteren.rs, cycle[h].rs, instret[h].rs, and update mod.rs

dkhayes117:
I'm not sure the set_clear_csr!() macros are correct in mcounteren.rs.  I get a build error that says `cannot find function `_clear` in this scope`


src/register/mcounteren.rs
```diff
@@ -0,0 +1,471 @@
...
+
+
+    read_csr_as!(Mcounteren, 0x306, __read_mcounteren);
+    write_csr!(0x306, __write_mcounteren);
+    set!(0x306, __set_mcounteren);
+    clear!(0x306, __clear_mcounteren);
+
+    set_clear_csr!(
+    /// User cycle Enable
+    , set_cy, clear_cy, 1 << 0);
```
dkhayes117, line 345: Ah, I think all of these should be set_mcounteren and clear_mcounteren

Disasm (changes requested):
In order to make this work, you will have to add a few assembly functions, namely `__read_mcounteren`, `__write_mcounteren`, `__set_mcounteren`, `__clear_mcounteren`. You can do this in `asm.S` with a single `RW(0x306, mcounteren)` line. Then you need to rebuild precompiled binaries with `assemble.sh`/`assemble.ps1` and commit the updated binaries.
Same goes for `cycle` and other CSRs you added. However, for them you can use `RO`/`RO32` macros.

src/register/cycle.rs
```diff
@@ -0,0 +1,6 @@
+//! cycle register
+//! Shadow of mcycle register that can be read in user mode
```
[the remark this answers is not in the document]
dkhayes117: I see, it should be pretty easy to add a scounteren.rs file and update comments to reflect this.

src/register/mcounteren.rs
```diff
@@ -0,0 +1,471 @@
...
+
+
+        /// User "hpm3" Enable
+        #[inline]
+        pub fn hpm3(&self) -> State {
+            match self.bits.get_bit(3) {
+                true => State::Enabled,
+                false => State::Disabled,
+            }
+        }
```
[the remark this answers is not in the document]
dkhayes117: That makes sense, and it makes the code less verbose.  index will need to be checked to make sure it is between 3 and 31.

src/register/mod.rs
```diff
@@ -14,87 +14,105 @@
...
 pub mod minstreth;
 
+
 // Machine Counter Setup
 mod mhpmeventx;
+
+
+
+
 pub use self::mhpmeventx::*;
```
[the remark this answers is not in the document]
dkhayes117: ran reformat command to clean this up I believe

dkhayes117:
> In order to make this work, you will have to add a few assembly functions, namely `__read_mcounteren`, `__write_mcounteren`, `__set_mcounteren`, `__clear_mcounteren`. You can do this in `asm.S` with a single `RW(0x306, mcounteren)` line. Then you need to rebuild precompiled binaries with `assemble.sh`/`assemble.ps1` and commit the updated binaries.
> Same goes for `cycle` and other CSRs you added. However, for them you can use `RO`/`RO32` macros.

I looked at the asm.S file and the `RW(0x306, mcounteren)` is already in there, along with the cycle, instret.... ones too.  The (3rd) commit I just made builds without errors

src/register/mcounteren.rs
```diff
@@ -0,0 +1,494 @@
...
+
+set_clear_csr!(
+/// User instret Enable
+, set_ir, clear_ir, 1 << 2);
+
+/*
+    set_clear_csr!(
+    /// User hpmcounter3 Enable
+    , set_hpm3, clear_hpm3, 1 << 3);
+
```
Disasm, line 379: Do you care to implement `set_hpm`/`clear_hpm` functions too?

src/register/mcounteren.rs
```diff
@@ -0,0 +1,494 @@
...
+    pub fn ir(&self) -> State {
+        match self.bits.get_bit(2) {
+            true => State::Enabled,
+            false => State::Disabled,
+        }
+    }
+
+    /// User "hpm[x]" Enable (bits 3-31)
+    #[inline]
+    pub fn hpm(&self, index: Index) -> State {
```
Disasm, line 86: I'm not sure you need this `Index` enum here as it's quite trivial. It can be replaced with usize and `assert!` bounds check.

src/register/cycle.rs
```diff
@@ -0,0 +1,7 @@
+//! cycle register
+//! Shadow of mcycle register
+//! must have mcounteren::cy bit enabled for use in supervisor mode (if implemented)
+//! if supervisor mode is not implemented this register will control user mode access
```
Disasm, line 4:
```suggestion
//! Must have `scounteren::cy` or `mcounteren::cy` bit enabled depending on whether
//! S-mode is implemented or not.
```
Maybe something like this?

Disasm (changes requested):
Sorry for the delay. A few notes:
* Lots of commented code, this should be removed before merge.
* Formatting check is still failing. Please run `cargo fmt`.

dkhayes117:
The set and clear functions in mcounteren.rs and scounteren.rs need to be looked at.  I don't understand the macro.rs file enough to know if they are correct.

src/register/mod.rs
```diff
@@ -61,13 +65,12 @@ pub mod mimpid;
 pub mod mvendorid;
 
 // Machine Trap Setup
-pub mod medeleg;
-pub mod mideleg;
```
Disasm, line 65: Why you removed these modules?

src/register/mod.rs
```diff
@@ -30,19 +30,23 @@ pub mod utval;
...
 // User Counter/Timers
-// TODO: cycle[h], instret[h]
-pub mod time;
-#[rustfmt::skip] // long macro use
+pub mod cycle;
 mod hpmcounterx;
+pub mod instret;
+pub mod time;
+
 pub use self::hpmcounterx::*;
```
Disasm, line 38:
```suggestion
mod hpmcounterx;
pub use self::hpmcounterx::*;
pub mod instret;
pub mod time;
```

It's strage to separate `mod` and `pub use` for the same module by several lines.

src/register/scounteren.rs
```diff
@@ -0,0 +1,65 @@
...
+/// User time Enable
+    , set_tm, clear_tm, 1 << 1);
+
+set_clear_csr!(
+/// User instret Enable
+    , set_ir, clear_ir, 1 << 2);
+
+#[inline]
+pub unsafe fn set_hpm(index: usize) {
+    assert!(((3..32).contains(&index)));
```
Disasm, line 57:
```suggestion
    assert!(3 <= index && index < 32);
```

src/register/scounteren.rs
```diff
@@ -0,0 +1,65 @@
...
+    _set(1 << index);
+}
+
+#[inline]
+pub unsafe fn clear_hpm(index: usize) {
+    assert!(((3..32).contains(&index)));
```
Disasm, line 63:
```suggestion
    assert!(3 <= index && index < 32);
```

src/register/scounteren.rs
```diff
@@ -0,0 +1,65 @@
...
+    /// User "instret[h]" Enable
+    #[inline]
+    pub fn ir(&self) -> bool {
+        self.bits.get_bit(2)
+    }
+
+    /// User "hpm[x]" Enable (bits 3-31)
+    #[inline]
+    pub fn hpm(&self, index: usize) -> bool {
+        assert!(((3..32).contains(&index)));
```
Disasm, line 33:
```suggestion
        assert!(3 <= index && index < 32);
```

src/register/mcounteren.rs
```diff
@@ -0,0 +1,65 @@
...
+    /// User "instret[h]" Enable
+    #[inline]
+    pub fn ir(&self) -> bool {
+        self.bits.get_bit(2)
+    }
+
+    /// User "hpm[x]" Enable (bits 3-31)
+    #[inline]
+    pub fn hpm(&self, index: usize) -> bool {
+        assert!(((3..32).contains(&index)));
```
Disasm, line 33:
```suggestion
        assert!(3 <= index && index < 32);
```

Disasm (changes requested):

src/register/mod.rs
dkhayes117: I think cargo fmt did that, I don't remember doing that.  Oh well, I fixed it :)

src/register/mod.rs
dkhayes117: I'm not sure how they got removed, my github fork had them there but my local repo had them as still //TODO. FIxed

Disasm:
Thanks! The only issue still left: `Mcounteren::cy()` and other `Mcounteren` method docs say that they enable access for User mode. However, if S-mode is implemented, they enable access to S-mode, not U-mode.

dkhayes117:
Appreciate your help and patience.  I might do another pull request after this one to look into setting the pmp registers without just writing a usize.  Something like enums for the addressing mode and permissions...I think even though the pmpcfg[x] regs are densely packed they are individually mapped.  What do you think?

Disasm (approved):
Thank you for your patience too!
Improving pmpcfg register handling sounds like a great idea!

PR #58, opened 2020-10-26
Add detail to pmpcfgx.rs to return the individual pmp[x]cfg register …

dkhayes117:
I wrote this about a dozen different ways.  Check it out and let me know what you think.  I didn't implement this for pmpcfg1 - pmpcfg3 as they will be identical setups as pmpcfg0 (well not quite identical).  Once I get pmpcfg0 satisfactory, I will add the rest to match.  Thanks,
Daniel


src/register/pmpcfgx.rs
```diff
@@ -1,7 +1,129 @@
...
+            match index {
+                0 => self.bits.get_bits(0..8),
+                1 => self.bits.get_bits(8..16),
+                2 => self.bits.get_bits(16..24),
+                3 => self.bits.get_bits(24..32),
+                4 => self.bits.get_bits(32..40),
+                5 => self.bits.get_bits(40..48),
+                6 => self.bits.get_bits(48..56),
+                7 => self.bits.get_bits(56..64),
+                _ => panic!(),
```
dkhayes117, line 52: If rv64 arch, pmp4cfg to pmp7cfg will be in pmpcfg0 register, I'm not sure how to check this.

src/register/pmpcfgx.rs
```diff
@@ -1,7 +1,129 @@
...
+    pub enum Permission {
+        NONE = 0,
+        R = 1,
+        W = 2,
+        RW = 3,
+        X = 4,
+        RX = 5,
+        WX = 6,
+        RWX = 7,
+    }
```
dkhayes117, line 16: It seems that having `X` only or `WX` may not make sense.  If you can write and execute from the address, why couldn't you read it too?  However, it seems to be possible to use those values, so I'm not sure.

Disasm: Let's keep them for completeness.

src/register/pmpcfgx.rs
Disasm: `#[cfg(riscv32)]`, `#[cfg(riscv64)]`

src/register/pmpcfgx.rs
```diff
@@ -1,7 +1,129 @@
...
+                1 => Range::TOR,
+                2 => Range::NA4,
+                3 => Range::NAPOT,
+                _ => panic!(),
+            }
+        }
+
+        #[inline]
+        fn permission(&self, byte: usize) -> Permission {
+            match byte.get_bits(0..4) {
```
Disasm, line 69: Probably `get_bits(0..3)` or return `Option<Permission>` with `None` if `A` is not set

src/register/pmpcfgx.rs
```diff
@@ -1,7 +1,129 @@
...
+    pub unsafe fn set_lock(index: usize) {
+        assert!(index < 8);
+        _set(1 << (7 + (index * 8)));
+    }
+
+    #[inline]
+    pub unsafe fn clear_lock(index: usize) {
+        assert!(index < 8);
+        _clear(1 << (7 + (index * 8)));
+    }
```
Disasm, line 126: We also need to handle `pmpcfgX` writes where X>0.

Disasm (commented):
Thank you for the PR!
First pack of comments blow

src/register/pmpcfgx.rs
dkhayes117: I added these to condition the assert! line, not sure that is correct.

src/register/pmpcfgx.rs
dkhayes117: sure

src/register/pmpcfgx.rs
```diff
@@ -1,7 +1,129 @@
...
+                0 => self.bits.get_bits(0..8),
+                1 => self.bits.get_bits(8..16),
+                2 => self.bits.get_bits(16..24),
+                3 => self.bits.get_bits(24..32),
+                4 => self.bits.get_bits(32..40),
+                5 => self.bits.get_bits(40..48),
+                6 => self.bits.get_bits(48..56),
+                7 => self.bits.get_bits(56..64),
+                _ => panic!(),
+            }
```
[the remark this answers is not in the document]
dkhayes117: Yeah that makes more sense and is much more concise

src/register/pmpcfgx.rs
```diff
@@ -1,7 +1,129 @@
...
+        }
+
+        #[inline]
+        fn range(&self, byte: usize) -> Range {
+            match byte.get_bits(4..6) {
+                0 => Range::OFF,
+                1 => Range::TOR,
+                2 => Range::NA4,
+                3 => Range::NAPOT,
+                _ => panic!(),
```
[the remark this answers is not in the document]
dkhayes117: replaced

src/register/pmpcfgx.rs
dkhayes117: Typo, changed to 0..3 and made Permission an Option

src/register/pmpcfgx.rs
dkhayes117: I have the original configurations for pmpcfg[x] where x>0, will they be duplicated code of pmpcfg0 or is there a way to combine them?  I do not know.

src/register/pmpcfgx.rs
```diff
@@ -1,7 +1,124 @@
...
+
+    #[derive(Clone, Copy, Debug)]
+    pub struct Pmpcfg0 {
+        bits: usize,
+    }
+
+    impl Pmpcfg0 {
+        ///Returns the pmp byte associated with the index
+        #[inline]
+        pub fn pmp_byte(&self, index: usize) -> usize {
```
Disasm, line 41:
```suggestion
        pub fn pmp_byte(&self, index: usize) -> u8 {
```
Probably since it returns byte, it should have `u8` return type?

src/register/pmpcfgx.rs
```diff
@@ -1,7 +1,124 @@
...
+                0 => Some(Permission::NONE),
+                1 => Some(Permission::R),
+                2 => Some(Permission::W),
+                3 => Some(Permission::RW),
+                4 => Some(Permission::X),
+                5 => Some(Permission::RX),
+                6 => Some(Permission::WX),
+                7 => Some(Permission::RWX),
+                _ => None,
+            }
```
Disasm, line 74: You should return `None` if `A` is set to `OFF`, but there is no check for the value of the `A` bit.

src/register/pmpcfgx.rs
```diff
@@ -1,7 +1,124 @@
...
+    }
+
+    read_csr_as!(Pmpcfg0, 0x3A0, __read_pmpcfg0);
+    write_csr!(0x3A0, __write_pmpcfg0);
+    set!(0x3A0, __set_pmpcfg0);
+    clear!(0x3A0, __clear_pmpcfg0);
+
+    #[inline]
+    pub unsafe fn set_permissions(permission: Permission, index: usize) {
+        assert!(index < 8);
```
Disasm, line 101: The upper bound in this and the following asserts is different for riscv32 and riscv64

src/register/pmpcfgx.rs
```diff
@@ -1,7 +1,124 @@
...
+        _set((permission as usize) << (index * 8));
+    }
+
+    #[inline]
+    pub unsafe fn set_range(range: Range, index: usize) {
+        assert!(index < 8);
+        _set((range as usize) << (3 + (index * 8)));
```
Disasm, line 108: It's a mistake. The `_set` function sets bits to 1, it doesn't "write" bits. It's ok to use it for separate bits, but not for bit fields.

Disasm (changes requested):
Sorry for the delay and thank you! Here is the second batch of comments.

Maybe we should improve API a little bit more. For example, to get PMP range one have to call `pmp_byte()`, get the raw value and then pass it to `range()`. It's strange, because `range` doesn't use `self` at all, it only uses the byte! I see two options how we can improve this:
1. Create another type (let's say `PmpByte(u8)`) and implement `range()` and `permission()` on top of it.
2. Make `pmp_byte` private and pass index to `range()` and `permission()` instead of a byte.

Also I noticed that there is no method to read the `locked` field.

As for the set_permissions/set_range methods, maybe it's a good idea to merge them into one. I don't know where you may need separate methods for setting up a single pmp entry.

dkhayes117:
I updated the pull request but I can't get it to pass the rustfmt test.  I've run `cargo fmt` on every level of the repo, not sure what the problem is, or what else to do.

almindor:
> I updated the pull request but I can't get it to pass the rustfmt test. I've run `cargo fmt` on every level of the repo, not sure what the problem is, or what else to do.

It seems there's a bug in the travis/CI script:

```
$ ci/script.sh
+'[' -n '' ']'
+'[' -n '' ']'
+'[' -n 1 ']'
+cargo fmt -- --check
Could not run rustfmt, please make sure it is in your PATH.
```
so unrelated to the PR. @Disasm might know what's going on.

Disasm:
Apparently, rustfmt started using the `RUSTFMT` env var and we set it to 1 to trigger the check. I fixed this in a separate PR.

dkhayes117:
Review please 🙏🙂

dkhayes117:
I added some .idea files from IntelliJ by accident, not sure how to remove them from my commit :(

Disasm:
@dkhayes117 you can `git reset --hard 06d429609f9c9648aa1706e5ba9b1d7ab125b519` and then `git push --force` to remove the last two commits from the PR.

dkhayes117:
@Disasm Appreciate the tip, looks like I got rid of a lot of unnecessary commits :+1:   Looks like it is ready for review at your convenience.

Disasm (commented):
Overall looks good to me, but it's an implementation for pmpcfg0, we still need implementations for other pmpcfgX. Maybe some of the structures can be moved to the parent module.

dkhayes117:
I get what you are saying but I am not sure about how to implement it. I moved what I thought would be common structs and enums in the latest commit, but not sure this is what you meant.  Let me know, thanks!

dkhayes117:
Yes! Got it to build :)

dkhayes117:
I probably can clean up the pmpcfg modules with a macro, but I haven't used them before.  Maybe I will give it a shot :)

src/register/pmpcfgx.rs
```diff
@@ -1,23 +1,350 @@
...
+    TOR = 1,
+    NA4 = 2,
+    NAPOT = 3,
+}
+
+/// PmpByte holds the a single pmp configuration
+#[derive(Clone, Copy, Debug)]
+pub struct PmpByte {
+    pub byte: u8,
+    //permission: Option<Permission>,
```
almindor, line 30: should these comments be removed?

almindor (commented):
Just a nit, otherwise I think this looks good.

src/register/pmpcfgx.rs
dkhayes117: Thank you for the review @almindor!  

I tested these changes out and there were some problems.  First off, the `set_permissions()` and `set_range()` functions used `_write` which caused the existing bits in the register to be overwritten.  I changed them to `_set` and all is well.  

Next, I found that the `get_range()` method was getting the wrong two bits, 4 and 5, it should be 3 and 4.  This was fixed and tested.  

I also removed the `clear_lock()` functions because they are pointless.  I forgot that once you set the lock on a pmp configuration, you cannot clear it.  It can only be cleared by power cycle.  

Finally, regarding the comments you mentioned on the `Pmpbyte` struct.  I don't like getting the byte from the Pmpcfg<x> struct, then after getting the byte having to call the individual methods for the configurations.  I would rather have `get_byte()` become `into_config()` and return a struct with the byte, range, permissions, and locked fields.  However, this will probably force the methods to be duplicated for each module instead of the single generic definition.  What do you think?

dkhayes117:
I wrote some sample code to test this PR on the HiFive1 Rev B.  
https://gist.github.com/dkhayes117/2e2b8d20ea20aa79cadaa96dbea8490a

almindor:
> I wrote some sample code to test this PR on the HiFive1 Rev B.
> https://gist.github.com/dkhayes117/2e2b8d20ea20aa79cadaa96dbea8490a

Perfect. Thanks. I'll try and get to it before next week.

dkhayes117:
PMP_Config

Cargo.toml
```diff
@@ -16,4 +16,4 @@ bit_field = "0.10.0"
 riscv-target = "0.1.2"
 
 [features]
-inline-asm = []
+inline-asm = []
```
almindor, line 19: why is this required?

src/register/pmpcfgx.rs
```diff
@@ -1,23 +1,293 @@
...
+        #[cfg(riscv64)]
+        assert!(index < 8);
+
+        _set(1 << (7 + index * 8));
+    }
 }
 
-/// Physical memory protection configuration, RV32 only
+/// Physical memory protection configuration
+/// Pmpcfg1 struct contains pmp4cfg - pmp7cfg for RV32 only
```
almindor, line 146: I wonder if we should limit this by #[cfg] for RV64 so it's completely inaccessible on that target.

almindor (changes requested):
Few more points to consider, looking good otherwise.

src/register/pmpcfgx.rs
dkhayes117: I was just thinking about this last night.  Totally agree.  I will make the change.

Cargo.toml
dkhayes117: I think that is an option in some of the macros

almindor (approved):
LGTM

almindor (approved):

PR #61, opened 2021-01-06
Use a different environment variable to run rustfmt check

Disasm:


luojia65:
Included in: https://github.com/rust-embedded/riscv/pull/65

PR #63, opened 2021-01-26
Fix unescaped "[]"s in doc comment

longfangsong:
When running `cargo doc` on this crate and crates depends on this crate, cargo will always raise `unresolved link` warnings for those `[h]` and `[x]`, which is annoying.

This pr escaped the `[` and `]` in these code to silence these warning.


Disasm (approved):
Thank you!
bors r+

PR #64, opened 2021-02-01
Add `nop` the asm module

longfangsong:
`NOP` is an important instruction to prevent delays from being optimized away. But currently there's no no wrapper function for it in the asm module.

This pr simply adds it.


Disasm:
One can use `core::sync::atomic::compiler_fence` instead.

Also `__nop` function doesn't exist in asm.S.

PR #67, opened 2021-03-08
Fix register configuration and documents

luojia65:
This pull request contains:

1. Use `#[cfg(target_pointer_width = "32")]` other than `#[cfg(riscv32)]` to provide support on IDE lint and cargo doc outputs
2. Add `#[inline]` on most of `set` and `write` functions

### DIFF
--- a/src/register/satp.rs
+++ b/src/register/satp.rs
@@ -1,6 +1,5 @@
 //! satp register
 
-#[cfg(riscv)]
 use bit_field::BitField;
 
 /// satp register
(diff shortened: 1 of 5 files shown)


src/register/satp.rs
```diff
@@ -42,55 +41,65 @@ impl Satp {
...
 }
 
-#[cfg(riscv64)]
+/// 64-bit satp mode
+#[cfg(target_pointer_width = "64")]
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub enum Mode {
+    /// No translation or protection
     Bare = 0,
+    /// Page-based 32-bit virtual addressing
```
Disasm, line 87:
```suggestion
    /// Page-based 39-bit virtual addressing
```

Disasm (changes requested):
Looks good, thank you!
Just one small typo to fix :)

src/register/satp.rs
luojia65: Thanks! :)

Disasm (approved):
Thanks!
bors r+

PR #68, opened 2021-03-09
Bug fix for interrupt bit in `scause::set`

luojia65:
This pull request includes a bug fix. On RISC-V's cause register, first bit means is this cause an interrupt. The code before made a mistake to or an one bit when exception. After this fix, it correctly or an one bit when interrupt.

Related to https://github.com/luojia65/rustsbi/issues/10. Thank you @wyfcyx


Disasm (approved):
Thank you for the fix!
bors r+

PR #70, opened 2021-05-02
Add docs.rs document build metadata for build targets

luojia65:
This pull request changes build targets when build documents on `docs.rs`, it will show RISC-V targets other than default x86 targets.

Before:

image

After (take `rustsbi` crate as example): 

image


Disasm (approved):
Thank you!
bors r+

PR #71, opened 2021-05-05
Replace llvm_asm! with new asm! syntax

remimimimi:
For motivation and disadvantages see [RFC#2873-inline-asm](https://github.com/rust-lang/rfcs/blob/master/text/2873-inline-asm.md)


Disasm:
Thank you for the PR! I think in the future we should switch to a better approach implemented in `cortex-m`: it uses an assembly library generated from Rust code with nightly compiler.

PR #72, opened 2021-05-27
feat: Add implementations of user-mode trap registers

duskmoon314:
This PR add methods to read/write user-mode trap registers similar to other modes.

- Add `sedeleg`
  - `bits`: read in bits
  - methods to read each bit
  - `set_clear_csr`
- Add `sideleg`
  - `bits`: read in bits
  - `usoft`: read whether UserSoftInterrupt is enable
  - `utimer`: read whether UserTimerInterrupt is enable
  - `uext`: read whether UserExternalInterrupt is enable
  - `set_clear_csr`
- Impl `sip`
  - `set_clear_csr`
- Impl `sstatus`
  - `set_xie`/`set_xpie`: methods to form the value that will be written to the register
- Impl `ucause`
  - enum `Trap`
  - enum `Interrupt`
  - enum `Exception`
  - `code` `cause` `is_interrupt` `is_exception` `set`: methods similar to `scause`
- `uip`
  - `set_cleear_csr`
- `ustatus`
  - `set_uie`/`set_upie`: methods to form the value that will be written to the register

I also add some simple docs of these registers.


src/register/sstatus.rs
```diff
@@ -94,6 +94,31 @@ impl Sstatus {
     pub fn sd(&self) -> bool {
         self.bits.get_bit(size_of::<usize>() * 8 - 1)
     }
+
+    #[inline]
+    pub fn set_spie(&mut self, val: bool) {
+        self.bits.set_bit(5, val);
+    }
```
luojia65, line 101: What can we benefit from these `set_*` functions?

duskmoon314: For now, I could only give an example of initialing the user mode program in rCore-Tutorial.

We use a struct `TrapContext` for trap handler and initialing the user mode program.

```rust
#[repr(C)]
pub struct TrapContext {
    pub x: [usize; 32],
    pub sstatus: Sstatus,
    pub sepc: usize,
    pub kernel_satp: usize,
    pub kernel_sp: usize,
    pub trap_handler: usize,
}
```

In this case, we form `sstatus` in this way:

```rust
// rcore-os/rCore-Tutorial-v3/os/src/trap/context.rs L23-L25
let mut sstatus = sstatus::read();
sstatus.set_spp(SPP::User);
```

[source code](https://github.com/rcore-os/rCore-Tutorial-v3/blob/81559f70b9bab318bf80ba06a9252985dff32f49/os/src/trap/context.rs#L23-L25)

We don't modify `sstatus` immediately. And `set_*` functions provide a readable construction method.

almindor: this is similar to the approach in [this PR](https://github.com/rust-embedded/riscv/pull/77)

We're currently debating if using a newtype separate from the main setter [similar to this description](https://github.com/rust-embedded/riscv/pull/77#issuecomment-886264605) would work better.

@disasm any thoughts on this? seems like having individual bitsetters is a recurring problem/solution

duskmoon314:
I noticed that the N extension has been temporarily moved out of the spec. So this PR can be closed at any time now.

PR #86, opened 2021-08-11
feat: Use new asm! instead of llvm_asm!

duskmoon314:
The new `asm` macro has been merged into `nightly` for some time, while `llvm_asm` will be gradually deprecated and will not enter `stable`. This PR replaces `llvm_asm` with `asm`, but there are still some minor issues to discuss.

My reference are:
- [asm feature](https://doc.rust-lang.org/unstable-book/library-features/asm.html)
- [arch/cortex-m: change a few llvm_asm!s to asm!s](https://github.com/tock/tock/pull/2092)
- [Convert all uses of llvm_asm! to asm!](https://github.com/rust-lang/stdarch/pull/1052)

---

In the new `asm` feature, there are several options can set to optimize codes. [options](https://doc.rust-lang.org/unstable-book/library-features/asm.html#options-1)

However, I'm not quite sure which options to set. (Sort of newbie to asm in rust)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Changed
+
+- Use new `asm!` instead of `llvm_asm!`
+
 ## [v0.7.0] - 2020-07-29
 
 ### Added
--- a/src/asm.rs
+++ b/src/asm.rs
@@ -7,7 +7,7 @@ macro_rules! instruction {
         pub unsafe fn $fnname() {
             match () {
                 #[cfg(all(riscv, feature = "inline-asm"))]
-                () => llvm_asm!($asm :::: "volatile"),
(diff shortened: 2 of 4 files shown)


Disasm:
Thank you! Could you also add a change log entry?

duskmoon314:
@Disasm ok, done

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Changed
+
+- - Use new `asm!` instead of `llvm_asm!`
```
dkhayes117, line 12: Do we need the double bullet points ?

dkhayes117 (commented):
Looks good to me overall.

CHANGELOG.md
duskmoon314: oh, my mistake (|| -_-)

Disasm (approved):

PR #91, opened 2022-01-08
Add inline-asm build to CI, fix build

Disasm:
https://github.com/rust-embedded/riscv/pull/86 introduced new `asm!` syntax which is enabled with the `inline-asm` feature. However, it doesn't work with current nightly since `asm_const` unstable feature is not enabled. This PR fixes the problem and adds automatic checks to CI.


dkhayes117 (approved):
LGTM

Disasm:
bors r=dkhayes117

PR #93, opened 2022-01-26
Remove riscv target post fix triple information from asm archive name

MabezDev:
For RISCV targets that don't have `unknown-none-elf` in their name. The only bit that matters is the first section anyways.


MabezDev:
Whoops! Forgot about that. Done :)

almindor (approved):
lgtm, thanks

PR #96, opened 2022-02-23
Update embedded-hal dependency to 1.0.0-alpha.7

ahmedcharles:


Disasm:
Thank you for the PR! Could you explain how these alpha-updates should work? Do we need to release an alpha version of the `riscv` crate with this change? In any case, I don't think it's a good idea to merge it directly to master as this would create problems for regular releases.

ahmedcharles:
My current goal is to update various dependencies across all the crates locally so I can at least try to test them. The PRs are mostly to avoid duplicate effort.

Once I can flash something to my hifive revb, I'll make more concrete suggestions for how to release.

Thanks for commenting.

ahmedcharles:
I updated the PR to update the version and switched to the staging branch. I think this would be the first crate that needs to be updated now that I have a test project compiling/working.

Disasm:
`staging` is a special branch used by bors for testing. I don't think it's a good idea to merge anything into it as it will be removed later.

ahmedcharles:
Ok, is there another branch I should merge it into for an alpha release?

romancardenas:
Closing, as this has been addressed in #147

PR #98, opened 2022-03-03
Add builders for registers

jwnhy:
This is a rework of #77. In #77, there is detailed explaination about why this is needed.
I have taken advices that we should add a new type to store values of registers like `mstatus`.
So I added `MstatusBuilder` and `MstatusValue` to resolve this issue.

- `MstatusBuilder` contains a set of setters that can construct a value for `mstatus` register.
- `MstatusValue` is only able to write to the physical register and take effect.

As this is a rather big change, I plan to open this PR as a **request for comments**.

If this style is okay, I will work in a few days implementing and documenting for other registers.

Thanks in advance for your valuable comments.


jwnhy:
I am thinking that for common registers, just `setter` and `getter` should suffice.
For complicated register like `pmpcfg`, a more approachable config function should be provided?

jwnhy:
Move to staging as there are quite a lot changes.

PR #99, opened 2022-03-19
Remove uses of unstable features and always use inline assembly

taiki-e:
Currently, the `inline-asm` feature uses unstable features, but by replacing the uses of asm_const with `concat!`+`stringify`, the uses of unstable features can be removed.

https://github.com/rust-embedded/riscv/blob/cd31989ba11d5d64e1addd8aab98bfb00dd927d5/src/lib.rs#L17-L18


src/register/macros.rs
```diff
@@ -1,13 +1,13 @@
 macro_rules! read_csr {
-    ($csr_number:expr, $asm_fn: ident) => {
+    ($csr_number:literal, $asm_fn: ident) => {
```
taiki-e, line 2: The change from `expr` to `literal` is to prevent expressions like `1<<2` (it will be stringified as "1<<2") from being accidentally accepted.

.github/workflows/ci.yaml
```diff
@@ -36,7 +36,8 @@ jobs:
         run: cargo check --target riscv32imac-unknown-none-elf
       - name: Run CI script for riscv32imac-unknown-none-elf (inline-asm) under ${{ matrix.rust }}
         run: |
-          if [ "${{ matrix.rust }}" == "nightly" ]; then
+          # asm! requires Rust 1.59
```
Disasm, line 39: I think it would be better to bump MSRV to 1.59 and remove the feature-gates.

taiki-e:
Removed inline-asm feature and updated MSRV to 1.59.

Disasm (approved):
Sorry for delay and thank you! Let's release this PR as part of v0.8.0

PR #105, opened 2022-07-30
Add extra documentation for `delay`

Joeyh021:
Took me a minute or two to figure out what value exactly to pass to `McycleDelay::new`, so figured I'd make a quick PR to add a note for anyone else that hits the same confusion as I did. Also added a doc line for the `delay` module while I was at it, as it was missing one.


almindor (approved):

almindor:
Thanks!

PR #106, opened 2022-07-31
Add embedded-hal v1.0.0-alpha.8 support

almindor:
Fixed `Delay` trait for use with embedded-hal `v1.0.0-alpha.8 `

### DIFF
--- a/src/delay.rs
+++ b/src/delay.rs
@@ -1,6 +1,7 @@
-//! Delay devices and providers
+use core::convert::Infallible;
+
 use crate::register::mcycle;
-use embedded_hal::blocking::delay::{DelayMs, DelayUs};
+use embedded_hal::delay::blocking::DelayUs;
 
 /// Machine mode cycle counter (`mcycle`) as a delay provider
 #[derive(Copy, Clone)]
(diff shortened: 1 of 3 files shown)


almindor:
@Disasm Not quite sure how we want to support pre-releases. I guess we could keep the PRs open and just push pre-release crates.io (I think that's how e-h did it?)

This is the first in a big list of deps to support all of:
https://github.com/rust-embedded/riscv-rt/pull/101
https://github.com/rust-embedded/riscv/pull/106
https://github.com/riscv-rust/e310x/pull/26
https://github.com/riscv-rust/e310x-hal/pull/45

Disasm:
I think only `riscv` and `e310x-hal` require update as we should be able to use multiple `riscv` versions simultaneously. I think it's ok to create a release from a PR branch if it works, but please don't merge these branches to `master` as it may cause problems in the future.

src/delay.rs
```diff
@@ -17,71 +18,16 @@ impl McycleDelay {
     }
 }
 
-impl DelayUs<u64> for McycleDelay {
+impl DelayUs for McycleDelay {
+    type Error = Infallible;
+
     #[inline]
-    fn delay_us(&mut self, us: u64) {
+    fn delay_us(&mut self, us: u32) -> Result<(), Self::Error> {
```
rmsyn, line 25: In the latest `embedded-hal` release candidate, [embedded-hal-1.0.0-alpha.11](https://docs.rs/embedded-hal/1.0.0-alpha.11/embedded_hal/delay/trait.DelayUs.html) this trait no longer returns a `Result`, nor has an associated `Error` type.

romancardenas: Shouldn't we aim at [1.0.0-rc.1](https://docs.rs/embedded-hal/1.0.0-rc.1/embedded_hal/index.html)?

rmsyn: > Shouldn't we aim at [1.0.0-rc.1](https://docs.rs/embedded-hal/1.0.0-rc.1/embedded_hal/index.html)?

Definitely, same story there: https://docs.rs/embedded-hal/1.0.0-rc.1/embedded_hal/delay/trait.DelayUs.html

Wasn't sure about the separation, so just went with the `alpha` line of releases.

romancardenas:
#147 points to `e-h-1.0.0-rc1` and solves all the conflicts. It also bumps MSRV to 1.60.

romancardenas:
Closing, as this has been addressed in #147

PR #107, opened 2022-08-07
fix reading marchid and mimpid

orangecms:
It seems to have been a copy-paste error.
We ran into a very nasty bug in oreboot on the Allwinner D1 (C906) where
an errata patch in Linux relies on those two being zero.

Signed-off-by: Daniel Maslowski <info@orangecms.org>

### DIFF
--- a/src/register/marchid.rs
+++ b/src/register/marchid.rs
@@ -16,7 +16,7 @@ impl Marchid {
     }
 }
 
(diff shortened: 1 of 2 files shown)


Disasm (approved):
Good catch, thank you!

luojia65:
Nice catch, will update RustSBI according to it

PR #108, opened 2022-08-14
fix: clearify that mip.{MSIP, MTIP} are read-only

### LINKED ISSUE #62: The MTIP bit in mip register is read-only
In the `riscv-privileged` manual, it is said that "The MTIP bit (in mip) is read-only and is cleared by writing to the memory-mapped machine-mode timer compare register."  
But in the `riscv` crate, there is an interface like this:  
```Rust
use riscv::register::mip;
use riscv_rt::entry;

#[entry]  
fn main() {
    mip::set_mtimer();
}
```
We can write the code `mip::set_mtimer()` to set the 7th bit, the MTIP bit, of the mip register, which does not permitted in `riscv-privileged` manual.  
Is there some reason for the existence of `mip::set_mtimer()`? If not, I suggest remove it.  
Thanks.

luojia65:
closes #62

In RISC-V privileged specification, it says:

> Bits mip.MTIP and mie.MTIE are the interrupt-pending and interrupt-enable bits for machine timer interrupts. MTIP is read-only in mip, and is cleared by writing to the memory-mapped machine-mode timer compare register.
>
> Bits mip.MSIP and mie.MSIE are the interrupt-pending and interrupt-enable bits for machine-level software interrupts. MSIP is read-only in mip, and is written by accesses to memory-mapped control registers, which are used by remote harts to provide machine-level interprocessor interrupts.

indicated by the specification, mip.MSIP and mip.MTIP bits are read-only. This pull request clearifies this by removing {set, clear}_{msoft, mtimer} functions from mip module of riscv crate.

### DIFF
--- a/src/register/mip.rs
+++ b/src/register/mip.rs
@@ -81,17 +81,11 @@ set_clear_csr!(
     /// Supervisor Software Interrupt Pending
     , set_ssoft, clear_ssoft, 1 << 1);
 set_clear_csr!(
-    /// Machine Software Interrupt Pending
-    , set_msoft, clear_msoft, 1 << 3);
-set_clear_csr!(
     /// User Timer Interrupt Pending
     , set_utimer, clear_utimer, 1 << 4);
 set_clear_csr!(
     /// Supervisor Timer Interrupt Pending
     , set_stimer, clear_stimer, 1 << 5);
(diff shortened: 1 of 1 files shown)


almindor (approved):
nice find. thanks!

PR #109, opened 2022-08-21
Update to edition 2021.

Dirbaio:
This shouldn't be a breaking change since Edition 2021 came out in
Rust 1.56, and MSRV is already higher than that (Rust 1.59)


dkhayes117 (approved):
Lgtm!

Disasm (approved):
Thank you!
bors r+

PR #110, opened 2022-08-21
Add critical-section 1.0 implementation, fix multicore unsoundness.

Dirbaio:
~~Requires #109~~

This adds a [critical-section](https://github.com/rust-embedded/critical-section) implementation for single-core chips, based on disabling all interrupts.

`interrupt::free` is is unsound on multicore systems because it only disables interrupts in the
current core. For multicore chips, a chip-specific critical section implementationis needed instead. Unsoundness is fixed by not returning the `CriticalSection` token.

This is a breaking change.

This is the riscv equivalent of https://github.com/rust-embedded/cortex-m/pull/447 and https://github.com/rust-embedded/cortex-m/pull/448

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,11 @@ targets = [
     "riscv64imac-unknown-none-elf", "riscv64gc-unknown-none-elf",
 ]
 
+[features]
+critical-section-single-hart = ["critical-section/restore-state-bool"]
+
 [dependencies]
 bare-metal = "1.0.0"
 bit_field = "0.10.0"
+critical-section = "1.1.0"
 embedded-hal = "0.2.6"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,6 @@ targets = [
 critical-section-single-hart = ["critical-section/restore-state-bool"]
 
 [dependencies]
-bare-metal = "1.0.0"
 bit_field = "0.10.0"
 critical-section = "1.1.0"
 embedded-hal = "0.2.6"
--- a/src/interrupt.rs
+++ b/src/interrupt.rs
@@ -2,9 +2,8 @@
 
 // NOTE: Adapted from cortex-m/src/interrupt.rs
 use crate::register::mstatus;
-pub use bare_metal::{CriticalSection, Mutex};
 
-/// Disables all interrupts
+/// Disables all interrupts in the current hart.
 #[inline]
 pub unsafe fn disable() {
     match () {
@@ -15,11 +14,11 @@ pub unsafe fn disable() {
     }
 }
 
-/// Enables all the interrupts
+/// Enables all the interrupts in the current hart.
 ///
 /// # Safety
 ///
-/// - Do not call this function inside an `interrupt::free` critical section
+/// - Do not call this function inside a critical section.
 #[inline]
 pub unsafe fn enable() {
     match () {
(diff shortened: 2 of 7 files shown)


dkhayes117 (commented):
Thank you!  Looks good. but I will defer to disasm for final review.

dkhayes117 (approved):

Cargo.toml
```diff
@@ -16,7 +17,11 @@ targets = [
     "riscv64imac-unknown-none-elf", "riscv64gc-unknown-none-elf",
 ]
 
+[features]
+critical-section-single-core = ["critical-section/restore-state-bool"]
+
 [dependencies]
 bare-metal = "1.0.0"
 bit_field = "0.10.0"
+critical-section = "1.1.0"
```
Disasm, line 26: Make it optional maybe?

Disasm:
> The goal is to release it in 0.8.x so critical-section becomes usable without having to wait for 0.9.

Next release will be "breaking" anyways because one breaking change is already merged. I'd prefer to have this change together with other critical-section-related changes.

Cargo.toml
Dirbaio: It's required unconditionally by the `singleton!` macro

(and the macro can still be used if `critical-section-single-core` is not enabled, if the user supplies a c-s implementation from somewhere else, such as a chip-specific impl for a multicore chip, from its HAL crate)

almindor:
@Dirbaio sorry I complicated this a bit, we released 0.9 due to a botched 0.8.1 (which contained some other breaking changes and we had to yank it). Could you please bump the changelog? I think that's the only blocker here.

Dirbaio:
Rebased

src/interrupt.rs
```diff
@@ -2,9 +2,8 @@
 
 // NOTE: Adapted from cortex-m/src/interrupt.rs
 use crate::register::mstatus;
-pub use bare_metal::{CriticalSection, Mutex};
 
-/// Disables all interrupts
+/// Disables all interrupts in the current core.
```
almindor, line 6: I wonder if we should specify what `core` means in RISC-V context here. e.g. is it a hard or an actual core with its own fetching unit.

almindor (commented):
LGTM, but I think we might want to look at `core` vs `hart` terminlogy and meaning before proceeding.

Dirbaio:
I'm afraid I don't know enough about riscv to know what's the difference between "core" and "hart" so I'll do whatever you guys choose :)

FawazTirmizi:
In this case, I think `hart` is the correct term to use over `core`. The `acquire()` and `release()` functions disable/enable interrupts for the hart by flipping `mstatus::mie`. 

> Interrupts seem to be mixed between these two, some seem to be core-wide and some seem to be hard-specific but maybe I'm misunderstanding this.

As far as I am aware, there is nothing ratified that defines "core-wide" interrupts, although an interrupt controller like the PLIC could be configured to route a set of interrupts to all of the harts on a core. This eventually may not always be true, as Section 2 of the [Smclic extension](https://github.com/riscv/riscv-fast-interrupt/blob/master/clic.adoc) does say:

>  While the current CLIC provides only hart-local interrupt control, future additions might also support directing interrupts to harts within a core, hence the name ...

However, there isn't any further mention of this in the spec so far.

It's entirely possible that I've missed something here; there are a _lot_ of specs relating to interrupts.

almindor:
> In this case, I think `hart` is the correct term to use over `core`. The `acquire()` and `release()` functions disable/enable interrupts for the hart by flipping `mstatus::mie`.
> 
> > Interrupts seem to be mixed between these two, some seem to be core-wide and some seem to be hard-specific but maybe I'm misunderstanding this.
> 
> As far as I am aware, there is nothing ratified that defines "core-wide" interrupts, although an interrupt controller like the PLIC could be configured to route a set of interrupts to all of the harts on a core. This eventually may not always be true, as Section 2 of the [Smclic extension](https://github.com/riscv/riscv-fast-interrupt/blob/master/clic.adoc) does say:
> 
> > While the current CLIC provides only hart-local interrupt control, future additions might also support directing interrupts to harts within a core, hence the name ...
> 
> However, there isn't any further mention of this in the spec so far.
> 
> It's entirely possible that I've missed something here; there are a _lot_ of specs relating to interrupts.

I think you are correct, we should probably rename things to be hart specific.

Dirbaio:
Changed `core` -> `hart`.

almindor (approved):
Thank you!

PR #111, opened 2022-08-21
Fix `interrupt::free()` unsoundness on multicore systems.

Dirbaio:
depends on #109, #110

This is unsound on multicore systems because it only disables interrupts in the
current core. For multicore chips, a chip-specific critical section implementation
is needed instead.

Unsoundness is fixed by not returning the `CriticalSection` token.

This is a breaking change.

This is the riscv equivalent of https://github.com/rust-embedded/cortex-m/pull/447

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -6,7 +6,10 @@
 /// at most once in the whole lifetime of the program.
 ///
 /// # Note
-/// this macro is unsound on multi-core systems
+///
+/// this macro requires a `critical-section` implementation to be set. For single core systems, you can
+/// enable the `critical-section-single-core` feature for this crate. For multi core systems, you
+/// have to provide one from elsewhere, typically your chip's HAL crate.
 ///
 /// # Example
 ///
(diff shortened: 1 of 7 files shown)


dkhayes117 (approved):
Again, lgtm.  Defer to disasm/almindor for final review.

src/macros.rs
```diff
@@ -6,7 +6,10 @@
 /// at most once in the whole lifetime of the program.
 ///
 /// # Note
-/// this macro is unsound on multi-core systems
+///
+/// this macro requires a `critical-section` implementation to be set. For single core systems, you can
```
Disasm, line 10: Maybe feature-gate this macro with `critical-section-single-core`?

Dirbaio: The macro can still be used if `critical-section-single-core` is not enabled, if the user supplies a c-s implementation from somewhere else (such as a chip-specific impl for a multicore chip, from its HAL crate).

Dirbaio:
Folded into #110

PR #112, opened 2022-10-06
Propose to release a version `riscv` v0.8.1

luojia65:
The `riscv` crate includes an important fix (https://github.com/rust-embedded/riscv/pull/107) where the RustSBI community would leverage the use of. However, the latest release v0.8.0 did not include this fix. In this pull request we want to propose to release a new version of `riscv` crate to include at least this fix, so that RustSBI community would write correct implemenations based on this crate.

Notes to `CHANGELOG.md`: the release date `2022-10-06` can always be changed.

r? @Disasm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [v0.8.1] - 2022-10-06
+
 ### Fixed
 
 - Fix `asm::delay()` to ensure count register is always reloaded
+- Fix reading marchid and mimpid (#107)
 
 ## [v0.8.0] - 2022-04-20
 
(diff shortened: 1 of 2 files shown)


almindor (approved):
LGTM

almindor:
@luojia65 0.8.1 is now released on crates.io

Disasm:
This should be v0.9.0 release since it introduces a breaking change: https://github.com/rust-embedded/riscv/commit/9418b6c59a9f2726dfc00e69af61834c32de6663

almindor:
> This should be v0.9.0 release since it introduces a breaking change: [9418b6c](https://github.com/rust-embedded/riscv/commit/9418b6c59a9f2726dfc00e69af61834c32de6663)

Hmm good point. Wasn't that broken to begin with though? In other words anyone who used those methods would just get an error?

FawazTirmizi:
`mip` is WARL, and as far as I can tell the spec doesn't say that an attempt to write to a read-only bit on these registers would cause an exception. I tested this on a HiFive Inventor and it seems to just ignore it, which is what I'd expect to happen in this case.

almindor:
> `mip` is WARL, and as far as I can tell the spec doesn't say that an attempt to write to a read-only bit on these registers would cause an exception. I tested this on a HiFive Inventor and it seems to just ignore it, which is what I'd expect to happen in this case.

Ah ok, let me punch this to 0.9 then and yank 0.8.1 @Disasm agreed?

Disasm:
@almindor sounds good, thank you!

almindor:
We should be good now

PR #117, opened 2022-10-16
fix atomicity of critical section, fixes #116

### LINKED ISSUE #116: The `critical_section` implementation is wrong
These two lines is not atomic and will cause problems

https://github.com/rust-embedded/riscv/blob/d9c60763b8fb53097ad1d95ff99c1eabe216dc34/src/critical_section.rs#L11-L13

and could be replaced by a single `csr` instruction:
```rust
let mut mstatus: usize;
asm!("csrrci {}, 0x300, 0b100", out(reg) mstatus);
core::mem::transmute::<_, Mstatus>(mstatus).mie()
```
which prevents from being interrupted between reading status and disabling interrupt.

almindor:
This changes `SingleHartCriticalSection::acquire` to be done in an atomic way.

Fixes #116

NOTE: no changelog since it fixes a small bug on unreleased change

### DIFF
--- a/src/critical_section.rs
+++ b/src/critical_section.rs
@@ -8,9 +8,9 @@ set_impl!(SingleHartCriticalSection);
 
 unsafe impl Impl for SingleHartCriticalSection {
     unsafe fn acquire() -> RawRestoreState {
-        let was_active = mstatus::read().mie();
-        interrupt::disable();
-        was_active
+        let mut mstatus: usize;
+        core::arch::asm!("csrrci {}, 0x300, 0b100", out(reg) mstatus);
+        core::mem::transmute::<_, mstatus::Mstatus>(mstatus).mie()
     }
 
     unsafe fn release(was_active: RawRestoreState) {


almindor:
Ugh sorry about the force push mess, I couldn't get the feature to register for some reason, always checks passed but it  just ignored this code.

We can't use the `Mstatus::bits` trick since it'd allow instantiation (and cought by the compiler too)

Dirbaio:
I don't think it's a soundness problem. If the race condition happens, the only effect is to undo your manual `interrupt::disable`. It'll never undo an `interrupt::disable` inside `critical_section::with`. Therefore, you can't abuse this to create unsoundness in abstractions based on critical_section.

It can only break abstractions based on doing this weird "disable but not reenable", which I honestly can't think of a sane reason you'd want to do that. And anyway, 

I think in general, docs on `interrupt::{disable,enable}` should have a note saying "disable+enable pairs must be properly nested". 

- This particular race can only happen if in an ISR handler you disable irqs and then return from the ISR handler without reenabling them. (which honestly I have no idea why'd you want to do that lol).
- Even if this race is fixed, if you have "not properly nested" `interrupt::disable` *inside* a critical section, it'll be undone at critical section exit.

Dirbaio:
To be clear, if riscv can "get status + disable irqs" atomically then we should *of course* use it  It's great that it can, cortex-m can't :sweat: 

My point was that even with this fix, `disable` without a paired `enable` is still broken, so abstractions relying on that are still broken, so this is not a soundness fix for them.

Dirbaio (approved):
bors r+

Dirbaio:
of course I don't have permission :D

PR #121, opened 2023-01-17
fix critial section implementation

tfx2001:
It should be `0b1000` rather than `0b100`. `Bit 2` is a reserved bit, and `Bit 3` is `MIE` that needs to be cleared.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


dkhayes117 (dismissed):
Yikes!  Thanks for bringing this up.

almindor (approved):
Nice find and fix thanks! Could you please add CHANGELOG entry?

tfx2001:
> Nice find and fix thanks! Could you please add CHANGELOG entry?

Done.

dkhayes117 (approved):
thanks

PR #123, opened 2023-02-19
Added clippy to the CI workflow

romancardenas:
I've been working with this crate for a while and noticed that it didn't include clippy as part of the CI process. This PR checks that clippy does not trigger any warning for new contributions. I also ran `cargo clippy --fix` to "fix" the crate.

If you think this is useful, I can adapt other repos (e.g., riscv-rt or e310x). Cheers!

### DIFF
--- a/.github/workflows/clippy.yaml
+++ b/.github/workflows/clippy.yaml
@@ -1,0 +1,36 @@
+name: Clippy
+
+on:
+  push:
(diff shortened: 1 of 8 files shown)


almindor:
I was also wondering about clippy. Thanks! Will review when I get a moment

almindor (approved):

Issue #124, opened 2023-02-20
Add "standard" peripherals


romancardenas:
I'm interested on porting RTIC to RISC-V platforms, and I was comparing the [`cortex-m`](https://github.com/rust-embedded/cortex-m) crate and this one.
The `cortex-m` crate contains a `peripheral` module for the FPU, MPU, or the NVIC. I thought that it would be worth it to add a `peripheral` module for common/standard peripherals such as the [(A)CLIC](https://github.com/riscv/riscv-aclint/blob/main/riscv-aclint.adoc) or the [PLIC](https://github.com/riscv/riscv-plic-spec/blob/master/riscv-plic.adoc). These modules would be quite useful for i) re-using code for those microcontrollers that implement the peripherals and ii) defining a common `riscv-rtic` crate for porting RTIC for RISC-V.

What do you think?

techmccat:
I was planning on writing some initial peripheral code for the CLIC, but since it's not standardized yet i'm not sure how to go about it.
Ideally this generic version should be implemented against the latest draft spec, but there's already hardware out implementing older versions of the spec (my BL602 dev board has a SiFive E24 core) which has incompatible register definitions.

romancardenas:
Good catch, not sure about how to proceed with this. I suggest you implement the peripheral for your board, so you can test it. Then we can use feature flags to select a pre-release CLIC or a standard CLIC once done. In any case, please consider using the same approach as in #135, which is quite different from the one currently used in `master` for PLIC.

techmccat:
Wrote some initial code and pushed it to https://github.com/techmccat/riscv/tree/sifive-clic

Needs some work on CLINT (in)compatibility and separating che common code

techmccat:
After getting around to porting bl602-hal to vectored mode I think the volatile-register approach might be better than the one used in #135, mainly because it allows to interact with the interrupt controller via associated functions, like how it's done in the cortex-m crate with the NVIC, without needing to keep a global CLIC struct or instantiating one whenever you need it.

romancardenas:
Hmm I see... the main issue with the volatile-register approach and the Rust ecosystem is that base addresses of peripherals are target-dependent. Therefore we need to deal with const generics and type aliases everywhere. I'm open to having a discussion about this and evaluating pros and cons, let me know your opinions!

almindor:
[...]

PR #125, opened 2023-02-23
Generic PLIC peripheral

romancardenas:
I developed a generic driver for interfacing PLIC peripherals, as mentioned in #124.

# References

For developing this peripheral, I got inspired by the following resources:
- The [PLIC specification](https://github.com/riscv/riscv-plic-spec/blob/master/riscv-plic.adoc)
- The PLIC implementation in the [`e310x`](https://github.com/riscv-rust/e310x/blob/master/src/common/plic.rs) and [`e310x-hal`](https://github.com/riscv-rust/e310x-hal/blob/master/src/core/plic.rs) crates.
- The driver for NVIC in [`cortex-m`](https://github.com/rust-embedded/cortex-m/blob/master/src/peripheral/nvic.rs) crate.

# Particularities

For atomic operations, I use the [`volatile-register`](https://github.com/japaric/volatile-register) crate (as done in `cortex-m`).

Unlike the NVIC for Cortex-M microcontrollers, the standard leaves the base address as "implementation-specific". Thus, my implementation uses a const generic to define the base address of the PLIC. In PACs, developers can do something like this:
```rust
pub type PLIC = riscv::peripheral::PLIC<BASE_ADDRESS>;
```

I decided to use associated functions whenever possible, so we don't need to steal/pass the control of the PLIC. However, those functions that are not atomic are defined as methods.

Finally, I followed the safety guidelines of the NVIC implementation, so those associated functions that modify priorities/thresholds are marked as `unsafe`.

Let me know what you think. If you prefer any other style, I'll happily modify my implementation.


dkhayes117:
Thank you for the PR.  I'll look it over as soon as I can.

dkhayes117:
Not all riscv targets will have a PLIC, should this be feature gated?

dkhayes117:
Although, having to supply a PLIC base address would kind of act like a feature gate 🤔

romancardenas:
I'm fine with both. I plan to develop other peripherals (e.g., CLIC and (A)CLINT) in the same fashion. From what I read, they all have an implementation-specific base address. Using feature gates to only import the peripherals implemented on each PAC sounds reasonable.

romancardenas:
Maybe we need to reconsider the concept of a context for the PLIC peripheral. These are my thoughts:
- The main purpose of the contexts is to provide a "safe", isolated interface to the interrupts for each HART.
- Probably, there will be as many contexts as HARTs in the platform.
- Interrupt priorities are shared among contexts
- Each context enables/disables the interrupt sources for their context.
- Each HART attends interrupts via its context.

So I was thinking that the `Context` trait may be a bit annoying to maintain. Maybe it would be more useful to have one `PLIC` per context, and the context number could be a const generic too. From what I saw, most boards only implement one context. What do you think?

romancardenas:
@dkhayes117 I developed an alternative implementation of the PLIC that stores the context as a const generic.
I personally like this approach more, because otherwise, I think that dealing with the context number would be non-ergonomic in practice.

I also added a feature gate for optionally adding the peripheral. Let me know what you think!

romancardenas:
I added additional methods and generics to the current traits. The reason for that is that I am currently working on a port of RTIC for the E310X microcontrollers. As part of this work, I identified nice-to-have features that will ease the integration of RTIC to any platform with PLIC.

src/peripheral/plic.rs
```diff
@@ -0,0 +1,230 @@
...
+/// Maximum number of words needed to represent interrupts with flags.
+const MAX_FLAGS_WORDS: usize = MAX_SOURCES / (u32::BITS as usize);
+/// Maximum number of contexts supported by the PLIC standard.
+const MAX_CONTEXTS: usize = 15_872;
+
+/// Register block.
+#[repr(C)]
+pub struct RegisterBlock {
+    /// `0x0000_0000..=0x0000_0FFC` - Interrupt Priority Register.
+    pub priority: [RW<u32>; MAX_SOURCES],
```
dkhayes117, line 19: Priority 0x0 is reserved, maybe a check should be added to prevent it being used?

dkhayes117 (commented):
Looks good.  I left one comment about validating that the source priority != 0.  @almindor @Disasm, It would be nice to have another pair of eyes on this if possible.  Thanks!

src/peripheral/plic.rs
romancardenas: According to the standard, "a priority value of 0 is reserved to mean "never interrupt" and effectively disables the interrupt". So I think that 0 is a valid priority that disables the interruption indirectly.

dkhayes117: I see, yep, I think you are right on this.

dkhayes117 (dismissed):
I think this is good to merge, but I'll give it a couple of days to see if the other members have any opinions.

romancardenas:
@dkhayes117 take a look at my latest commit. Basically, I added utility functions to access the MIE and MIP CSRs. I also take advantage of the `try_from` methods of the traits to avoid returning raw numbers. This is aligned with the PLIC implementation of [`e310x-hal`](https://github.com/riscv-rust/e310x-hal/blob/master/src/core/plic.rs).

romancardenas:
Sorry for all these changes, but I've been discovering some deficiencies in my proposal.

Unlike the `cortex-m` crate, we need to provide a `new`/`default` constructor for the peripherals, as they are all optional and PACs are responsible for creating them when needed.

I developed an alternative version of the [`e310x`](https://github.com/greenlsi/e310x) crate that uses this generic implementation and compiles OK. Take a look to see how this peripheral would work.

I think this PR is **now** good to go (fingers crossed).

romancardenas:
Yaaay it works :D
I'm having some issues with the RTC and debugging, but the PLIC works as expected.

https://github.com/romancardenas/hifive1/blob/master/examples/interrupt.rs

romancardenas:
I opened a new PR with a clean commit history of this very same implementation

PR #126, opened 2023-02-26
lib: gate delay module and asm under `delay` feature

luojia65:
RISC-V Privileged Specification and Unprivileged Specification didn't really define a standard way of delaying a period of time. We gate delay features in `riscv` crate under `delay` feature, so users may enable this feature to use those custom features.

r? @dkhayes117


Disasm:
I don't get the point. Why this feature should be put behind a feature-gate? Does it introduce an incompatibility with some hardware implementation?

luojia65:
Thanks! Closing for

Issue #133, opened 2023-05-10
Replace bors with GitHub merge queues


romancardenas:
As mentioned in the [WG repository](https://github.com/rust-embedded/wg/issues/671), bors is about to stop working. The embedded WG suggested migrating to GitHub merge queues, and this repo could be an excellent candidate to test how this works.

#132 already updates the current CI workflows and is ready to support merge queues. However, we still need to activate the branch protection rule on the repo and, if everything goes well, remove bors. @rust-embedded/riscv are you OK with this?

romancardenas:
Right now, my PR should work for both bors and merge-queue

almindor:
merged + ticked the queue in settings

PR #132, opened 2023-05-08
Update GitHub workflows

romancardenas:
With all the bors thing going on, I've been reading a bit about merge queues etc. and noticed that some of our workflows still use outdated/deprecated actions. This PR fixes this.

PR #134, opened 2023-05-16
Remove bors

romancardenas:
Once we have the CI workflows updated and ready for GHMQ... it's time to remove all the bors-related files and configurations. Once this PR is accepted, bors will no longer be required and we can remove it from the repo settings once and for all.

Related PRs and issues: #132 #133 
Check also [this issue](https://github.com/rust-embedded/wg/issues/671) of the wg repo.

### DIFF
--- a/.github/bors.toml
+++ b/.github/bors.toml
@@ -1,8 +1,0 @@
-block_labels = ["needs-decision"]
-delete_merged_branches = true
-required_approvals = 1
-status = [
-    "build-check",
-    "clippy-check",
-    "rustfmt",
-]


romancardenas:
bors try

Urhengulas:
As far as I understand you can also remove the `push` event from the workflows:
https://github.com/rust-embedded/riscv/blob/bba4de26f16f3bf2a9358400a571946729fa3576/.github/workflows/build.yaml#L2-L3
Maybe you want to keep master, but the other two were just for bors, weren't they?

almindor:
> You're right! I'd leave it for `master`, but the other branches can be removed. @almindor are we missing something?

I think we should be good here.

PR #139, opened 2023-08-11
Rework of mcause and scause registers

romancardenas:
I did a rework of `mcause` and `scause` registers to follow the [latest RISC-V specification](https://github.com/riscv/riscv-isa-manual/releases/download/riscv-isa-release-1239329-2023-05-23/riscv-privileged.pdf). Namely, I removed the user-level interrupt sources and fixed a missing exception type in `scause`.

I also used the `From` and `TryFrom` traits, so now the `Interrupt` and `Exception` enumerations are more usable. I plan to re-export these enums in `riscv-rt` to avoid code duplicates and simplify a bit the logic of the runtime.


almindor (approved):
LGTM, thanks!

PR #140, opened 2023-08-14
Remove bit_field dependency

romancardenas:
I replaced the `bit_field` crate with bit-wise operations. In this way, we avoid needless checks for getting bits from the registers. Now, future versions of the `riscv` crate will be slightly faster (and smaller)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ Cargo.lock
 target/
 
 .vscode/
+.DS_Store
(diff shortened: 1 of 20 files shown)


romancardenas:
ping @rust-embedded/risc-v !

If you prefer, I can add private getter and setter methods to the register macros to make the code look "cleaner", but I think it is OK as is.

romancardenas:
ping @rust-embedded/riscv @rust-embedded/risc-v

almindor:
Hey sorry for the delay, I've been really busy lately. I'll try to review this ASAP.

almindor (approved):
LGTM

PR #141, opened 2023-09-05
asm: set unsafety of `nop` and `delay` functions to safe

luojia65:
The NOP instruction does not change architecturally visible state, including operations that would break memory safety boundaries; thus `nop` instruction wrapper should be safe to call. The `delay` function uses ADDI and BNE instructions which would form a function internal loop, which is safe under Rust constraints.

This pull request also includes an explanation on NOP instruction under Rust docs, and a small codestyle fix in Cargo.toml file.

I modified internal macro `instruction` to allow wrapping instruction with safe functions. If it's inappropriate, please reply and I'll change to other ways of implementation.


romancardenas (approved):
LGTM! I'll leave a couple of days for more experienced folks of the @rust-embedded/risc-v team to review it before merging it. Thanks!

PR #142, opened 2023-09-20
Rework on misa and mstatus

romancardenas:
Added new fields to `mstatus` register according to the latest RISC-V specification. For RISCV-32, some of the new fields are in a new register, `mstatush`.


almindor (approved):
LGTM, thank you

PR #143, opened 2023-09-21
Removing user mode registers

romancardenas:
Yet another small step to align this crate with the latest RISC-V specification.

I think I already removed all user mode-related stuff. If you detect that I missed something, please let me know and I'll remove it once and for all.


romancardenas:
ping @rust-embedded/riscv

MabezDev:
Could you point me to where this change is mentioned in the RISCV spec (and which document version)? I can't seem to find it, so I must be looking in the wrong place.

romancardenas:
This is the latest RISC-V priviledged specification: https://github.com/riscv/riscv-isa-manual/releases/download/Priv-v1.12/riscv-privileged-20211203.pdf

You can check in the preface that the N extension has been removed an is no longer part of the document.

almindor (approved):

almindor:
LGTM
NOTE: this forced a minor version bump since it's removal

almindor:
I'll let the next person merge this I think one more set of eyes would be good to ensure nothign was removed that shouldn't be.

romancardenas:
Great, thanks!

I'd like to remove all the peripheral-related stuff since I'm using riscv-peripheral from now on, review all the code as is, and publish a new version in crates.io. Perhaps we could also include the embedded-hal 1.0 implementation in this release.

MabezDev (approved):
Yeah I was looking at the wrong document :sweat_smile:, thanks for the link!

LGTM!

PR #147, opened 2023-10-16
Update embedded-hal to 1.0

romancardenas:
Preparing to support `e-h-1` once we get the first stable version.

Related: #96 and #106

### DIFF
--- a/src/delay.rs
+++ b/src/delay.rs
@@ -1,87 +1,30 @@
 //! Delay devices and providers
 use crate::register::mcycle;
-use embedded_hal::blocking::delay::{DelayMs, DelayUs};
+use embedded_hal::delay::DelayUs;
 
 /// Machine mode cycle counter (`mcycle`) as a delay provider
 #[derive(Copy, Clone)]
+#[repr(transparent)]
 pub struct McycleDelay {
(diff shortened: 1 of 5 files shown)


almindor (approved):

src/delay.rs
```diff
@@ -1,87 +1,30 @@
 //! Delay devices and providers
 use crate::register::mcycle;
-use embedded_hal::blocking::delay::{DelayMs, DelayUs};
+use embedded_hal::delay::DelayUs;
 
 /// Machine mode cycle counter (`mcycle`) as a delay provider
 #[derive(Copy, Clone)]
+#[repr(transparent)]
```
MabezDev, line 7: What's the purpose of `#[repr(transparent)]` here?

romancardenas: Not sure, to be honest. As far as I know, it is desirable for FFI with C to avoid potential weird compiler allocations. But maybe it is useless, I can remove it.

MabezDev: I don't think it is going to cause any harm lets leave it as is.

MabezDev (approved):
LGTM, thanks

PR #150, opened 2023-11-08
Transition to cargo workspace

romancardenas:
I've been working on adding new functionalities to the RISC-V environment, especially for standard peripherals. Additionally, it seems that the [`cortex-m` folks](https://github.com/rust-embedded/cortex-m/pull/488) are working on splitting the interrupt-related stuff from `cortex-m` to a new crate to ease the integration of breaking changes. I think it is a good idea to start using their approach, so we can maintain new RISC-V crates without having to create new repos, setup the CI, etc.

Let me know what you think about this structure and, if you are happy with it, I will move `riscv-rt` to this repo.


romancardenas:
ping @rust-embedded/riscv

almindor (approved):

almindor:
LGTM, we should've workspaced a lot of things ages ago

PR #159, opened 2023-11-29
Update for embedded-hal to 1.0.0-rc.2

purplethunder:
embedded-hal 1.0.0-rc.2 introduced [backwards incompatible changes](https://github.com/rust-embedded/embedded-hal/commit/4ec58c85d81a191a0efddcf7269d0a4c753b2b45) that break riscv builds.  This change pins the crate to 1.0.0-rc.1.

```
error[E0432]: unresolved import `embedded_hal::delay::DelayUs`
 --> riscv/src/delay.rs:3:5
  |
3 | use embedded_hal::delay::DelayUs;
  |     ^^^^^^^^^^^^^^^^^^^^^-------
  |     |                    |
  |     |                    help: a similar name exists in the module: `DelayNs`
  |     no `DelayUs` in `delay`

For more information about this error, try `rustc --explain E0432`.
```

### DIFF
--- a/riscv/src/delay.rs
+++ b/riscv/src/delay.rs
@@ -28,8 +28,8 @@ impl DelayNs for McycleDelay {
     #[inline]
    fn delay_ns(&mut self, ns: u32) {
         let t0 = mcycle::read64();
-        let us_64: u64 = ns.into();
-        let clock = (us_64 * (self.ticks_second as u64)) / 1_000_000_000u64;
+        let ns_64: u64 = ns.into();
+        let clock = (ns_64 * (self.ticks_second as u64)) / 1_000_000_000u64;
         while mcycle::read64().wrapping_sub(t0) <= clock {}
    }
 }
--- a/riscv/src/delay.rs
+++ b/riscv/src/delay.rs
@@ -1,6 +1,6 @@
 //! Delay devices and providers
 use crate::register::mcycle;
-use embedded_hal::delay::DelayUs;
+use embedded_hal::delay::DelayNs;
 
 /// Machine mode cycle counter (`mcycle`) as a delay provider
 #[derive(Copy, Clone)]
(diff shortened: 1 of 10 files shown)


MabezDev:
I don't think we should accept this PR, instead we should upgrade riscv to the newest release candidate. Would you mind PRing that instead?

riscv/src/delay.rs
```diff
@@ -19,12 +19,17 @@ impl McycleDelay {
...
     fn delay_us(&mut self, us: u32) {
+	self.delay_ns(us * 1000)
+    }
+
+    #[inline]
+   fn delay_ns(&mut self, ns: u32) {
         let t0 = mcycle::read64();
-        let us_64: u64 = us.into();
-        let clock = (us_64 * (self.ticks_second as u64)) / 1_000_000u64;
+        let us_64: u64 = ns.into();
```
romancardenas: Please, rename this to `ns_64`

riscv/src/delay.rs
```diff
@@ -19,12 +19,17 @@ impl McycleDelay {
     }
 }
 
-impl DelayUs for McycleDelay {
+impl DelayNs for McycleDelay {
     #[inline]
     fn delay_us(&mut self, us: u32) {
```
romancardenas, line 24: Can't we use the default implementation of this method?

romancardenas (changes requested):
Looks good to me!! I left you a couple of comments, please address them before accepting the PR

romancardenas:
Hmm... maybe we need to bump riscv-rt to 1.60?

romancardenas:
You will probably also need to update the CI to make it pass: https://github.com/rust-embedded/riscv/blob/f8c39238c815af9ca92ce945c0dc4ab83c58a004/.github/workflows/riscv-rt.yaml#L14

romancardenas:
Seems that the issue comes with `riscv-target`, used for building the `riscv-rt` crate and check the different RISC-V extensions...

I'll try to develop a quick solution to stop using this dependency and allow us to leave the MSRV to 1.60

romancardenas:
Closing as #160 now solves this

PR #160, opened 2023-12-02
`riscv-rt`: remove riscv-target

romancardenas:
I implemented the `riscv-target` necessary logic in `build.rs`. This way we avoid extra dependencies that triggered problems with the MSRV (see #159).


romancardenas:
This PR now solves the issues of #159

almindor (approved):

PR #161, opened 2023-12-04
`riscv-rt`: Rework on exception and interrupt handling

romancardenas:
This PR consists of three major contributions:
- Removed U-mode interrupts, as they are no longer in the standard
- Transition from `Vector` unions to `Option<fn>`. Both are equivalent, as `Option<fn>` exploits a niche optimization that uses `0usize` to represent `None`. However, using `Option`s is more idiomatic than `Union`s
- Added a mechanism to deal with exceptions that is similar to how we treat interrupts. Now, it allow us to define custom exception handlers for standard exception sources.

### DIFF
--- a/.github/workflows/riscv-rt.yaml
+++ b/.github/workflows/riscv-rt.yaml
@@ -1,6 +1,6 @@
 on:
   push:
-    branches: [ master ]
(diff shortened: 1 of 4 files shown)


almindor (approved):
LGTM

PR #164, opened 2023-12-07
Add `riscv-peripheral` crate

romancardenas:
Crate for standard RISC-V peripherals.

I've been using these peripheral APIs myself on an E310x and work properly.
Let me know if you want any extra change before merging it to master


riscv-peripheral/src/hal_async/aclint.rs
```diff
@@ -0,0 +1,49 @@
...
+    }
+}
+
+impl Future for DelayAsync {
+    type Output = ();
+
+    #[inline]
+    fn poll(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Self::Output> {
+        match self.mtime.read().wrapping_sub(self.t0) < self.n_ticks {
+            true => Poll::Pending,
```
Dirbaio, line 31: on `Pending` you should store the waker somewhere, and wake it when the timer expires. Otherwise the future will hang (except with bery dumb `block_on`-like executors that poll in a hot loop, not waiting for wakes)

romancardenas: Thanks for the comment! I left asynchronous stuff for later once Rust 1.75 was out. I'll address it ASAP

romancardenas: Hey, @Dirbaio I address your comments.

I'm pretty new to asynchronous programming, but if I'm not wrong this implementation is just a *soft* polling, as the executor must poll from time to time the `Future`. I was wondering about how an interrupt-driven approach would look like.

I guess that I would need a `static` placeholder to store all the pending contexts and wake them when the `MachineTimer` interrupt is triggered. However, it seems quite difficult to dimension this, as I don't know the number of contexts that may live concurrently in an asynchronous application. Can you send me a few pointers to get more insights about this?

romancardenas: So... `async` stuff was more scary than I expected 😮‍💨 

My latest push provides a few `extern "Rust"` functions that  must  be implemented for pushing/popping `Timer`s to a timer queue. If I'm not wrong, my approach would be compatible with embassy-time or any other interface by properly implementing the `_riscv_peripheral*` functions.

romancardenas:
I reimplemented the asynchronous delay for the (A)CLINT peripheral. Now, it relies on three `extern "Rust" fn _riscv_peripheral_aclint_*` functions that allows us to push/pop wakers to/from a timer queue such as the one provided by embassy-time.

romancardenas:
ping @rust-embedded/riscv

almindor (approved):
I like this approach, LGTM.

PR #169, opened 2023-12-28
`riscv`: ECALL and nested interrupts

romancardenas:
This PR adds a wrapper to trigger environment calls. It also adds a new function for allowing nested interrupts in M and S mode. Nested interrupts are specially useful for RTOSs, as they can implement preemption in this way.


almindor (approved):
LGTM! This makes preemption so much nicer, thanks!

almindor (approved):
Hah the variable == module wording got me a bit confused for  a bit. LTGM

PR #170, opened 2023-12-31
add `riscv-semihosting`

romancardenas:
I adapted [`riscv-semihosting`](https://github.com/riscv-rust/riscv-semihosting) to the current fashion and added it to the workspace.

The idea is that, by default, it assumes we run in machine mode (the same as in riscv and riscv-rt). The u-mode feature is used to cause exceptions if machine mode is not accesible.

I'm currently using it in my rtic branch for testing in QEMU and works as expected.


almindor (approved):
Very nice thank you! I  can only optimistically review atm. though. It'd be great if someone else could double check this all works.

MabezDev:
Is there a reason to not use the [semihosting](https://crates.io/crates/semihosting) crate? It looks like it already supports RISCV, along with other arches.

romancardenas:
To be honest, I didn't know this crate existed. It looks more complete than this PR, and it works perfectly in my rtic branch.

The only reason I would accept this PR is to provide the exact same interface as [`cortex-m-semihosting`](https://crates.io/crates/cortex-m-semihosting), which seems to be regularly used in the books, examples etc. The `semihosting` crate has numerous feature gates to enable more semihosting functionalities depending on your use case.

What do you think about maintaining `riscv-semihosting` to mimic the `cortex-m` environment but leaving a section in the `README.md` file stating that the [`semihosting`](https://crates.io/crates/semihosting) crate is way more complete and preferred for more professional development?

riscv-semihosting/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+            asm!("
+                .option push
+                .option norvc
+                slli x0, x0, 0x1f
+                ebreak
+                srai x0, x0, 0x7
+                .option pop
+            ",
+            inout("a0") nr,
+            in("a1") _arg,
```
taiki-e, line 232: There appear to be two bugs in this asm call.

First, [riscv-semihosting-spec](https://github.com/riscv-software-src/riscv-semihosting/blob/f2e0eaf46d34a54ebbc1799c9a8879c4cc544fe4/riscv-semihosting-spec.adoc#21-semihosting-trap-instruction-sequence) says:

> If paging is in use in the current mode, this sequence must not cross a page boundary as the semihosting system must be able to check for the semihosting sequence without needing data from potentially missing pages.

I believe you have to use `.balign 16`.

Second, some syscalls such as [`SYS_ELAPSED`](https://github.com/ARM-software/abi-aa/blob/HEAD/semihosting/semihosting.rst#633return) change `a1`, so I believe you have to use `inout` instead of `in` for `a1`.

See also my [semihosting crate](https://github.com/taiki-e/semihosting/blob/969d2350d51c38740649837f09e23a5a0de64cb8/src/sys/arm_compat/syscall/riscv.rs#L12-L25) for an example of RISC-V semihosting syscalls in Rust.

PR #171, opened 2024-01-02
`riscv-semihosting`: bug fixes

romancardenas:
This PR adds a reference to [`semihosting`](https://crates.io/crates/semihosting) in the `README.md` file, as this crate is preferred over `riscv-semihosting`.

I also addressed @taiki-e 's comments on #170 . (@taiki-e , could you take a look to confirm the issues are fixed?)

### DIFF
--- a/riscv-semihosting/src/lib.rs
+++ b/riscv-semihosting/src/lib.rs
@@ -231,7 +231,7 @@ pub unsafe fn syscall1(_nr: usize, _arg: usize) -> usize {
                 .option pop
             ",
             inout("a0") nr,
-            inout("a1") arg,
+            inout("a1") arg => _,
             options(nostack, preserves_flags),
             );
             nr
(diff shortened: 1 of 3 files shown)


riscv-semihosting/src/lib.rs
```diff
@@ -229,7 +231,8 @@ pub unsafe fn syscall1(_nr: usize, _arg: usize) -> usize {
                 .option pop
             ",
             inout("a0") nr,
-            in("a1") _arg,
+            inout("a1") arg,
```
taiki-e, line 234:
```suggestion
            inout("a1") arg => _,
```

The resulting value is not used, so currently it gets a warning.

https://github.com/rust-embedded/riscv/actions/runs/7388098852/job/20098197550

```
warning: value assigned to `arg` is never read
   --> riscv-semihosting/src/lib.rs:234:25
    |
234 |             inout("a1") arg,
    |                         ^^^
    |
    = help: maybe it is overwritten before being read?
    = note: `#[warn(unused_assignments)]` on by default
```

taiki-e (approved):

MabezDev (approved):

Issue #182, opened 2024-02-14
`riscv-rt`: nightly builds fail with `single-hart` feature: linker needs `default_mp_hook` symbol


romancardenas:
Looks now nightly now fails to compile `single-hart` examples, as `_mp_hook` is not defined and `default_mp_hook` is feature-gated. I guess we could just remove the feature gate, `default_mp_hook` is a very lightweight function.

PR #183, opened 2024-02-14
`riscv-rt`: Fix nightly builds

romancardenas:
Added an implementation of `default_mp_hook` even when `single-hart` feature is enabled. This is required to avoid linker errors in Nightly builds.


romancardenas:
Closes #182

jessebraham (approved):
LGTM!

PR #188, opened 2024-02-23
`riscv-rt`: add `pre_init_trap` handler to detect errors during the boot process

### LINKED ISSUE #156: Pre initialization trap handling
At the moment, the trap handler is initialized rather late in the boot procedure (it is more or less the last thing to happen before jumping to `main`). However, traps may happen before this, and they will not be handled appropriately (as far as I can tell, my CPU lands in a reset loop).

The most prominent situation in which this may happen is when I accidentally run the software compiled with multiplication instructions on a machine without them, because of these lines: https://github.com/rust-embedded/riscv-rt/blob/master/asm.S#L99-L110

I see multiple different approaches to improve the current situation:

- Simply hardcode the non-mul branch in the `asm.S`. Since the factor against we are multiplying is rather low (it's the number of harts of the machine), the run time overhead shouldn't hit that hard.
- Going one step further, I suggest compiling `asm.S` for the `rv32i` and `rv64i` targets only. Most of these extensions aren't needed anyways during initialization.
- Alternatively, define a pre-Rust trap handler which is initialized first in `_start`. Once the Rust runtime is ready, `default_setup_interrupts` will override it to the actual implementation.

romancardenas:
I'm working on enabling the vectored interrupt mode in `riscv-rt`. To do so, I'm following [SiFive's RISC-V interrupt Cookbook](https://starfivetech.com/uploads/sifive-interrupt-cookbook-v1p2.pdf) and the [`esp-riscv-rt`](https://github.com/esp-rs/esp-hal/blob/main/esp-riscv-rt/src/lib.rs) crate.

Section 2.1.3 of the SiFive Cookbook provides information about requiring a pre-init trap handler to detect errors before the boot process. This issue is also outlined in #156.

This PR adds a simple `pre_init_trap` routine in the `.text.abort` section. At the beginning of `_abs_start`, interrupts are disabled and the `xtvec` register is configured to handle exceptions or interrupts at `pre_init_trap`. Once the boot process is done, `_setup_interrupts` redirects interrupts to `_start_trap`.

I thought about leaving `pre_init_trap` user-defined with a dummy default implementation. However, it is quite dangerous to leave users to add their code before the boot process is done. Thus, in my opinion, it is better to leave it like this, being the new trap just kind of a flag to allow us to detect bugs in `riscv-rt`.

**BEWARE:** I'm far from an assembly expert, so a double-check is welcome 😁. Also, the folks from `esp-rs` (ping @jessebraham @MabezDev) could provide very valuable feedback on this.

Closes #156

### DIFF
--- a/riscv-rt/CHANGELOG.md
+++ b/riscv-rt/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- Add `pre_init_trap` to detect early errors during the boot process.
+
 ### Changed
 
 - Moved all the assembly code to `asm.rs`
--- a/riscv-rt/src/asm.rs
+++ b/riscv-rt/src/asm.rs
@@ -66,12 +66,19 @@ _abs_start:
     .option norelax
     .cfi_startproc
     .cfi_undefined ra",
+    // Disable interrupts
     #[cfg(feature = "s-mode")]
     "csrw sie, 0
     csrw sip, 0",
     #[cfg(not(feature = "s-mode"))]
     "csrw mie, 0
     csrw mip, 0",
+    // Set pre-init trap vector
+    "la t0, pre_init_trap",
+    #[cfg(feature = "s-mode")]
+    "csrw stvec, t0",
+    #[cfg(not(feature = "s-mode"))]
+    "csrw mtvec, t0",
 );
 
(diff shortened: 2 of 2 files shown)


almindor (approved):
LGTM, thanks!

romancardenas:
I could slightly modify the PR and define `_pre_init_trap` as a weak symbol so users could overwrite it. In any case, this should be properly documented.

PR #195, opened 2024-03-15
`riscv-peripheral`: Support Core-Local Interrupt Controller (CLIC) RISC-V Privileged Architecture Extensions

hegza:
Specification: <https://github.com/riscv/riscv-fast-interrupt/blob/master/clic.adoc>

We've made a research prototype hardware implementation of rv32emc + CLIC and have completed preliminary tests showing that this software implementation for CLIC works. However, there's a few open issues:

- How to deal with dynamic RISC-V registers in software?
    - `riscv::register::mcause` dynamically allocates a different number of bits for `code` depending on whether `riscv::mtvec::trap_mode` is set to `Clic = 0b11` or not.
    - The number of bits allocated for priority and level in `riscv_peripheral::clic::intctl` depends on the value set in `riscv_peripheral::clic::smclicconfig::mnlbits`.
- We're investigating whether the fact that unaligned writes to CLIC registers fail is a problem with our HW design or a problem with RISC-V spec. This is making the current code cleanliness of the `clicintx` registers slightly un-ideal.
- CLIC is not yet ratified, as is the custom around here in the RISC-V world :)
- The CLIC spec if pretty big so we've only implemented the parts that are relevant to our use case with RTIC

### DIFF
--- a/riscv-peripheral/CHANGELOG.md
+++ b/riscv-peripheral/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- Add support for CLIC <https://github.com/riscv/riscv-fast-interrupt/blob/master/clic.adoc>
+
 ## [v0.1.0] - 2024-02-15
 
 ### Added
--- a/riscv-peripheral/Cargo.toml
+++ b/riscv-peripheral/Cargo.toml
@@ -8,7 +8,7 @@ authors = ["The RISC-V Team <risc-v@teams.rust-embedded.org>"]
 categories = ["embedded", "hardware-support", "no-std"]
 description = "Interfaces for standard RISC-V peripherals"
 documentation = "https://docs.rs/riscv-peripheral"
-keywords = ["riscv", "peripheral", "clint", "plic"]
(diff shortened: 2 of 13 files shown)


romancardenas:
Wooohoo! Love to see this, thanks so much :) I'll take a look to your implementation ASAP.

perlindgren:
Beware that the specification of the CLIC is in flux. So what's in the current spec may change.

hegza:
Ah, the CLIC spec has actually just a few days ago left it up to the implementor to define the MMIO layout that this PR implements. The few platforms that do implement CLIC do use this layout however so I guess it can be useful 😃 Gosh. Let's sit on it a while and see where this goes.

romancardenas:
Forgot about this PR!

I'm OK with merging in master an implementation that matches the current specification and updating this implementation as the specification changes.

hegza:
OK yeah, that sounds reasonable. I've been writing a journal the last 2 weeks and I was just getting back into working on the real-time system implementing this CLIC starting halfway through next week. I think I can get some update done on this within the next month.

hegza:
Closed due to MMIO interface obsolescene. I'll re-open / link to this one once I've got the new CSR API working.

Issue #146, opened 2023-10-09
RFC: Platform-specific exception codes


SoulSharer:
![image](https://github.com/rust-embedded/riscv/assets/7026184/ac3de159-6a68-40d7-8b68-911cdd2fb74b)

Are there any plans to provide support for platform specific extensions in `riscv` crate?
Reading through *RISC-V Privileged Architectures* specification I see that there are cases where, for example, a `3.1.15 Machine Cause Register` has exception codes allocated for platform and custom use.

Was wondering if it would be possible to extend enums `Interrupt` and `Exception` from outside of the `riscv` crate itself.
The best I could came up thus far is adding "PlatformSpecific1-32" fields to `Interrupt` and creating a user-defined enum `PlatformInterrupt` with conversion between (`TryFrom<Interrupt> for PlaftormInterrupt`) where applicable. (same goes for `Exception`)

romancardenas:
Hi! I don't think we have plans to add this functionality right now. If you think this feature is useful, feel free to open a PR/RFC and we can discuss it further :)

In my opinion, we should look at how `cortex-m` deals with platform-specific interrupts and follow their approach.

romancardenas:
While working on #200, I've been thinking on how to provide this functionality. I propose the following approach:

- Remove weak symbols in linker file (i.e., the `PROVIDE` directives) for interrupt and exception sources
- Provide procedural macros for defining the interrupt and exception sources of a platform. These macros would be in charge of generating all the symbols related to interrupts and exceptions (e.g., `_start_trap_rust`, `__INTERRUPTS`, `__EXCEPTIONS`, `_vector_table`...). Additionally, the code will provide weak symbols to all these interrupts using global inline assembly.
- By default, `riscv-rt` will use this new macro to generate a code that would be equivalent to its current version. I expect this change not to be a breaking change.
- A new feature would be able to opt-out the standard implementation. In this way, a new `my-riscv-specific-rt` runtime crate would have `riscv-rt` as a dependency, re-exporting everything and calling to the interrupt macro accordingly to the exception codes of the target.

PR #200, opened 2024-04-11
`riscv-rt`: Support for vectored mode interrupt handling

### LINKED ISSUE #158: Implement support for hardware interrupt dispatching
Currently there is a single interrupt entry point. RISCV CPU's also have a hardware dispatching mode.

romancardenas:
I am working on providing support for interrupt dispatching in vectored mode. This is a first draft, and I could not test yet if it works. However, I wanted to share it with you so I can start getting some feedback and suggestions about different design decisions. So, the main changes are:

- New `v-trap` feature to enable vectored mode. If this feature is enabled, the assembly code will add a trap vector with `j` instructions to a different trap per interrupt source. Some targets do not implement the standard interrupt convention (e.g., ESP32C3). That is why I left the table as weak. PACs will be able to replace this table accordingly.
- New `interrupt` procedural macro to implement interrupt handlers. Currently, in direct mode, this macro is quite dummy. However, I plan to add some parameters to the macro that will, eventually, allow us to "check" at compile time that the interrupt source name is an interrupt source of the target. Additionally, if `v-trap` feature is enabled, this macro will generate the trap handler to make sure that user registers are properly restored after executing the interrupt handler.

Also, I've been careful regarding how traps are generated to ease shortly the compatibility with the E extension (see #178).

This PR is still WIP, and I appreciate the feedback. Especially from the esp-rs folks (@MabezDev , @jessebraham) and @hegza , who already implemented a solution for this.

Closes #158

### DIFF
--- a/riscv-rt/macros/src/lib.rs
+++ b/riscv-rt/macros/src/lib.rs
@@ -356,15 +356,14 @@ fn store_trap<T: FnMut(&str) -> bool>(arch: RiscvArch, mut filter: T) -> String 
         RiscvArch::Rv32 => (4, "sw"),
         RiscvArch::Rv64 => (8, "sd"),
     };
-    let mut stores = Vec::new();
-    for (i, reg) in TRAP_FRAME
+
+    TRAP_FRAME
         .iter()
         .enumerate()
         .filter(|(_, &reg)| filter(reg))
-    {
-        stores.push(format!("{store} {reg}, {i}*{width}(sp)"));
-    }
-    stores.join("\n")
+        .map(|(i, reg)| format!("{store} {reg}, {i}*{width}(sp)"))
+        .collect::<Vec<_>>()
+        .join("\n")
 }
 
 /// Generate the assembly instructions to load the trap frame.
@@ -374,11 +373,12 @@ fn load_trap(arch: RiscvArch) -> String {
         RiscvArch::Rv32 => (4, "lw"),
         RiscvArch::Rv64 => (8, "ld"),
     };
-    let mut loads = Vec::new();
-    for (i, reg) in TRAP_FRAME.iter().enumerate() {
-        loads.push(format!("{load} {reg}, {i}*{width}(sp)"));
-    }
-    loads.join("\n")
+    TRAP_FRAME
+        .iter()
+        .enumerate()
+        .map(|(i, reg)| format!("{load} {reg}, {i}*{width}(sp)"))
+        .collect::<Vec<_>>()
+        .join("\n")
 }
 
 /// Generates weak `_start_trap` function in assembly for RISCV-32 targets.
@@ -428,7 +428,7 @@ fn weak_start_trap(arch: RiscvArch) -> TokenStream {
     format!(
         r#"
 core::arch::global_asm!(
-".section .trap, \\"ax\\"
+".section .trap, \"ax\"
 .align {width}
 .weak _start_trap
 _start_trap:
@@ -481,7 +481,7 @@ fn vectored_interrupt_trap(arch: RiscvArch) -> TokenStream {
     let instructions = format!(
         r#"
 core::arch::global_asm!(
-".section .trap, \\"ax\\"
+".section .trap, \"ax\"
 
 .global _start_DefaultHandler_trap
 _start_DefaultHandler_trap:
@@ -586,7 +586,7 @@ fn start_interrupt_trap(ident: &syn::Ident, arch: RiscvArch) -> proc_macro2::Tok
     let instructions = format!(
         r#"
 core::arch::global_asm!(
-    ".section .trap, \\"ax\\"
+    ".section .trap, \"ax\"
     .align 2
     .global _start_{interrupt}_trap
     _start_{interrupt}_trap:
--- a/riscv-rt/macros/src/lib.rs
+++ b/riscv-rt/macros/src/lib.rs
@@ -312,3 +312,136 @@ pub fn loop_global_asm(input: TokenStream) -> TokenStream {
     let res = format!("core::arch::global_asm!(\n\"{}\"\n);", instructions);
     res.parse().unwrap()
 }
+
+enum RiscvArch {
+    Rv32,
+    Rv64,
+}
+
+#[proc_macro_attribute]
+pub fn interrupt_riscv32(args: TokenStream, input: TokenStream) -> TokenStream {
+    interrupt(args, input, RiscvArch::Rv32)
+}
+
+#[proc_macro_attribute]
+pub fn interrupt_riscv64(args: TokenStream, input: TokenStream) -> TokenStream {
+    interrupt(args, input, RiscvArch::Rv64)
+}
+
+fn interrupt(args: TokenStream, input: TokenStream, _arch: RiscvArch) -> TokenStream {
+    let f = parse_macro_input!(input as ItemFn);
+
+    // check the function signature
+    let valid_signature = f.sig.constness.is_none()
+        && f.sig.asyncness.is_none()
+        && f.vis == Visibility::Inherited
+        && f.sig.abi.is_none()
+        && f.sig.generics.params.is_empty()
+        && f.sig.generics.where_clause.is_none()
+        && f.sig.variadic.is_none()
+        && match f.sig.output {
+            ReturnType::Default => true,
+            ReturnType::Type(_, ref ty) => matches!(**ty, Type::Never(_)),
+        };
+
+    if !valid_signature {
+        return parse::Error::new(
+            f.span(),
+            "`#[interrupt]` function must have signature `[unsafe] fn() [-> !]`",
+        )
+        .to_compile_error()
+        .into();
+    }
+
+    if !args.is_empty() {
+        return parse::Error::new(Span::call_site(), "This attribute accepts no arguments")
+            .to_compile_error()
+            .into();
+    }
+
+    // XXX should we blacklist other attributes?
+    let attrs = f.attrs;
+    let ident = f.sig.ident;
+    let block = f.block;
+
+    #[cfg(not(feature = "v-trap"))]
+    let start_trap = proc_macro2::TokenStream::new();
+    #[cfg(feature = "v-trap")]
+    let start_trap = v_trap::start_interrupt_trap_asm(&ident, _arch);
+
+    quote!(
+        #start_trap
+        #[export_name = #ident]
+        #(#attrs)*
+        pub unsafe fn #ident() #block
+    )
+    .into()
+}
+
+#[cfg(feature = "v-trap")]
+mod v_trap {
+    use super::*;
+
+    const TRAP_SIZE: usize = 16;
+
+    #[rustfmt::skip]
+    const TRAP_FRAME: [&str; TRAP_SIZE] = [
+        "ra",
+        "t0",
+        "t1",
+        "t2",
+        "t3",
+        "t4",
+        "t5",
+        "t6",
+        "a0",
+        "a1",
+        "a2",
+        "a3",
+        "a4",
+        "a5",
+        "a6",
+        "a7",
+    ];
+
+    pub fn start_interrupt_trap_asm(
+        ident: &syn::Ident,
+        arch: RiscvArch,
+    ) -> proc_macro2::TokenStream {
+        let function = ident.to_string();
+        let (width, store, load) = match arch {
+            RiscvArch::Rv32 => (4, "sw", "lw"),
+            RiscvArch::Rv64 => (8, "sd", "ld"),
+        };
+
+        let (mut stores, mut loads) = (Vec::new(), Vec::new());
+        for (i, r) in TRAP_FRAME.iter().enumerate() {
+            stores.push(format!("        {store} {r}, {i}*{width}(sp)"));
+            loads.push(format!("        {load} {r}, {i}*{width}(sp)"));
+        }
+        let store = stores.join("\n");
+        let load = loads.join("\n");
+
+        #[cfg(feature = "s-mode")]
+        let ret = "sret";
+        #[cfg(not(feature = "s-mode"))]
+        let ret = "mret";
+
+        let instructions = format!(
+            "
+core::arch::global_asm!(
+    \".section .trap, \\\"ax\\\"
+    .align {width}
+    _start_{function}_trap:
+        addi sp, sp, - {TRAP_SIZE} * {width}
+{store}
+        call {function}
+{load}
+        addi sp, sp, {TRAP_SIZE} * {width}
+        {ret}\"
+);"
+        );
+
+        instructions.parse().unwrap()
+    }
+}
--- a/riscv-rt/macros/src/lib.rs
+++ b/riscv-rt/macros/src/lib.rs
@@ -313,9 +313,120 @@ pub fn loop_global_asm(input: TokenStream) -> TokenStream {
     res.parse().unwrap()
 }
 
+#[derive(Clone, Copy)]
 enum RiscvArch {
     Rv32,
     Rv64,
+}
+
+const TRAP_SIZE: usize = 16;
+
+#[rustfmt::skip]
+const TRAP_FRAME: [&str; TRAP_SIZE] = [
+    "ra",
+    "t0",
+    "t1",
+    "t2",
+    "t3",
+    "t4",
+    "t5",
+    "t6",
+    "a0",
+    "a1",
+    "a2",
+    "a3",
+    "a4",
+    "a5",
+    "a6",
+    "a7",
+];
+
+fn store_trap<T: FnMut(&str) -> bool>(arch: RiscvArch, mut filter: T) -> String {
+    let (width, store) = match arch {
+        RiscvArch::Rv32 => (4, "sw"),
+        RiscvArch::Rv64 => (8, "sd"),
+    };
+    let mut stores = Vec::new();
+    for (i, reg) in TRAP_FRAME
+        .iter()
+        .enumerate()
+        .filter(|(_, &reg)| filter(reg))
+    {
+        stores.push(format!("{store} {reg}, {i}*{width}(sp)"));
+    }
+    stores.join("\n")
+}
+
+fn load_trap(arch: RiscvArch) -> String {
+    let (width, load) = match arch {
+        RiscvArch::Rv32 => (4, "lw"),
+        RiscvArch::Rv64 => (8, "ld"),
+    };
+    let mut loads = Vec::new();
+    for (i, reg) in TRAP_FRAME.iter().enumerate() {
+        loads.push(format!("{load} {reg}, {i}*{width}(sp)"));
+    }
+    loads.join("\n")
+}
+
+#[proc_macro]
+pub fn weak_start_trap_riscv32(_input: TokenStream) -> TokenStream {
+    weak_start_trap(RiscvArch::Rv32)
+}
+
+#[proc_macro]
+pub fn weak_start_trap_riscv64(_input: TokenStream) -> TokenStream {
+    weak_start_trap(RiscvArch::Rv64)
+}
+
+fn weak_start_trap(arch: RiscvArch) -> TokenStream {
+    let width = match arch {
+        RiscvArch::Rv32 => 4,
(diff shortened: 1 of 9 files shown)


hegza:
For sure I can at least test this on our HW, compare it with mine and see how it fits into the current workflow. I'll be onto this.

romancardenas:
I successfully ran an example on QEMU using `MachineTimer` and `MachineSoft` interrupts in vectored mode.

For each interrupt source, an independent trap handler is added in the final binary:

```asm
20010150 <_start_MachineSoft_trap>:
20010150:      	addi	sp, sp, -64
20010152:      	sw	ra, 0(sp)
20010154:      	sw	t0, 4(sp)
20010156:      	sw	t1, 8(sp)
20010158:      	sw	t2, 12(sp)
2001015a:      	sw	t3, 16(sp)
2001015c:      	sw	t4, 20(sp)
2001015e:      	sw	t5, 24(sp)
20010160:      	sw	t6, 28(sp)
20010162:      	sw	a0, 32(sp)
20010164:      	sw	a1, 36(sp)
20010166:      	sw	a2, 40(sp)
20010168:      	sw	a3, 44(sp)
2001016a:      	sw	a4, 48(sp)
2001016c:      	sw	a5, 52(sp)
2001016e:      	sw	a6, 56(sp)
20010170:      	sw	a7, 60(sp)
20010172:      	auipc	ra, 0
20010176:      	jalr	446(ra) // resulting from call MachineSoft
2001017a:      	lw	ra, 0(sp)
2001017c:      	lw	t0, 4(sp)
2001017e:      	lw	t1, 8(sp)
20010180:      	lw	t2, 12(sp)
20010182:      	lw	t3, 16(sp)
20010184:      	lw	t4, 20(sp)
20010186:      	lw	t5, 24(sp)
20010188:      	lw	t6, 28(sp)
2001018a:      	lw	a0, 32(sp)
2001018c:      	lw	a1, 36(sp)
2001018e:      	lw	a2, 40(sp)
20010190:      	lw	a3, 44(sp)
20010192:      	lw	a4, 48(sp)
20010194:      	lw	a5, 52(sp)
20010196:      	lw	a6, 56(sp)
20010198:      	lw	a7, 60(sp)
2001019a:      	addi	sp, sp, 64
2001019c:      	mret

200101a0 <_start_MachineTimer_trap>:
200101a0:      	addi	sp, sp, -64
200101a2:      	sw	ra, 0(sp)
200101a4:      	sw	t0, 4(sp)
200101a6:      	sw	t1, 8(sp)
200101a8:      	sw	t2, 12(sp)
200101aa:      	sw	t3, 16(sp)
200101ac:      	sw	t4, 20(sp)
200101ae:      	sw	t5, 24(sp)
200101b0:      	sw	t6, 28(sp)
200101b2:      	sw	a0, 32(sp)
200101b4:      	sw	a1, 36(sp)
200101b6:      	sw	a2, 40(sp)
200101b8:      	sw	a3, 44(sp)
200101ba:      	sw	a4, 48(sp)
200101bc:      	sw	a5, 52(sp)
200101be:      	sw	a6, 56(sp)
200101c0:      	sw	a7, 60(sp)
200101c2:      	auipc	ra, 0
200101c6:      	jalr	392(ra)
200101ca:      	lw	ra, 0(sp)
200101cc:      	lw	t0, 4(sp)
200101ce:      	lw	t1, 8(sp)
200101d0:      	lw	t2, 12(sp)
200101d2:      	lw	t3, 16(sp)
200101d4:      	lw	t4, 20(sp)
200101d6:      	lw	t5, 24(sp)
200101d8:      	lw	t6, 28(sp)
200101da:      	lw	a0, 32(sp)
200101dc:      	lw	a1, 36(sp)
200101de:      	lw	a2, 40(sp)
200101e0:      	lw	a3, 44(sp)
200101e2:      	lw	a4, 48(sp)
200101e4:      	lw	a5, 52(sp)
200101e6:      	lw	a6, 56(sp)
200101e8:      	lw	a7, 60(sp)
200101ea:      	addi	sp, sp, 64
200101ec:      	mret
```

We also preserve the original `_start_trap`, as it is the default implementation for all those interrupts that are not defined using the macro:
```asm
20010200 <_start_trap>:
20010200:      	addi	sp, sp, -64
20010204:      	sw	ra, 0(sp)
20010208:      	sw	t0, 4(sp)
2001020c:      	sw	t1, 8(sp)
20010210:      	sw	t2, 12(sp)
20010214:      	sw	t3, 16(sp)
20010218:      	sw	t4, 20(sp)
2001021c:      	sw	t5, 24(sp)
20010220:      	sw	t6, 28(sp)
20010224:      	sw	a0, 32(sp)
20010228:      	sw	a1, 36(sp)
2001022c:      	sw	a2, 40(sp)
20010230:      	sw	a3, 44(sp)
20010234:      	sw	a4, 48(sp)
20010238:      	sw	a5, 52(sp)
2001023c:      	sw	a6, 56(sp)
20010240:      	sw	a7, 60(sp)
20010244:      	add	a0, sp, zero // main difference with interrupt traps, as we need to forward the PC
20010248:      	jal	0x2001042a <_start_trap_rust>
2001024c:      	lw	ra, 0(sp)
20010250:      	lw	t0, 4(sp)
20010254:      	lw	t1, 8(sp)
20010258:      	lw	t2, 12(sp)
2001025c:      	lw	t3, 16(sp)
20010260:      	lw	t4, 20(sp)
20010264:      	lw	t5, 24(sp)
20010268:      	lw	t6, 28(sp)
2001026c:      	lw	a0, 32(sp)
20010270:      	lw	a1, 36(sp)
20010274:      	lw	a2, 40(sp)
20010278:      	lw	a3, 44(sp)
2001027c:      	lw	a4, 48(sp)
20010280:      	lw	a5, 52(sp)
20010284:      	lw	a6, 56(sp)
20010288:      	lw	a7, 60(sp)
2001028c:      	addi	sp, sp, 64
20010290:      	mret
```

Next, the vector table is filled as expected:
```asm
20010300 <_vector_table>:
20010300:      	j	0x20010200 <_start_trap>
20010304:      	j	0x20010200 <_start_trap>
20010308:      	j	0x20010200 <_start_trap>
2001030c:      	j	0x20010150 <_start_MachineSoft_trap>
20010310:      	j	0x20010200 <_start_trap>
20010314:      	j	0x20010200 <_start_trap>
20010318:      	j	0x20010200 <_start_trap>
2001031c:      	j	0x200101a0 <_start_MachineTimer_trap>
20010320:      	j	0x20010200 <_start_trap>
20010324:      	j	0x20010200 <_start_trap>
20010328:      	j	0x20010200 <_start_trap>
2001032c:      	j	0x20010200 <_start_trap>
```

I used GDB to monitor which trap handler is used and works as expected. The program does not use the `_start_trap` handler at all :D

riscv-rt/macros/src/lib.rs
```diff
@@ -312,3 +312,251 @@ pub fn loop_global_asm(input: TokenStream) -> TokenStream {
...
+        #[cfg(feature = "s-mode")]
+        let ret = "sret";
+        #[cfg(not(feature = "s-mode"))]
+        let ret = "mret";
+
+        let instructions = format!(
+            "
+core::arch::global_asm!(
+    \".section .trap, \\\"ax\\\"
+    .align {width} // TODO is this necessary?
```
hegza, line 549: In my understanding, we only need `_continue_interrupt_trap` to be a valid target for `j` instruction. `j` has LSB set to 0 so we'd need to make sure this is `align(2)` even if there's multiple items placed in `.trap` section. I can't see why it would ever need to be 4 or 8 byte aligned. I could be wrong.

hegza:
The proc-macro trap handler is somewhat more intimidating than the previous `trap_handler` macro. I guess it becomes worth it as the vectored handler needs a more customizable code flow anyway.

hegza:
Works as expected on Renode simulated generic 64-bit RISC-V using a CLINT for interrupt dispatch. I tested `MachineSoft` & `MachineTimer` as well.

I wonder how it meshes with the nested interrupts we use on the FPGA implementation of the real-time RVE. I'll try adapting that runtime to this one and see if it raises any issue.

romancardenas:
I tested this with my SLIC crate, which exploits nested `MachineSoft` interrupts to have an interrupt-driven program (e.g., RTIC). So hopefully you face no issues.

romancardenas:
**TO DO LIST**

- [ ] I would love to skip `_start_trap` at all and leave it only for exceptions. We could use the `DefaultHandler` to do this: https://github.com/rust-embedded/riscv/blob/aca641040c5bd3ef856bee94f2db0e235d6bb317/riscv-rt/src/lib.rs#L478)
- [ ] use cargo flags to simplify `_start_rust_trap` if vectored mode is enabled? maybe the array of `__INTERRUPTS` is not necessary in this mode, as only exceptions are expected to fall here.

Next, another challenge is supporting heterogeneous interrupt sources (e.g., target-specific additional interrupt sources or non-standard interrupt sources. In order to achieve this, I would do as follows:

- [ ] Modify `_start_rust_trap`. When an interrupt is detected, it relies on `_handle_interrupts` function, which, by default, behaves as [this](https://github.com/rust-embedded/riscv/blob/aca641040c5bd3ef856bee94f2db0e235d6bb317/riscv-rt/src/lib.rs#L496) part.
- [ ] Provide a new macro to 1) define core interrupt enums 2) define the appropriate `_vector_table`

Essentially, with these changes, we would start moving towards supporting target-specific interrupt sources (see #146)

romancardenas:
New changes to `riscv-rt`:

- Now, when an interrupt is detected, `_start_trap_rust` calls `_dispatch_interrupt(interrupt_number)`. This does not really change anything from previous versions, but we are now closer to accepting custom interrupt sources (I will leave this to a future PR, it needs additional changes).
- When `v-trap` feature is enabled, the binary will contain a `_vector_table` with `j` instructions to trap handlers. The first `j` instruction points to `_start_trap`, as it is reserved to exceptions. The rest points to `_start_INTERRUPT_trap`, where `INTERRUPT` is the ID of the corresponding interrupt name (e.g., `MachineSoft`). By default, all `_start_INTERRUPT_trap` are weak aliases of `_start_DefaultHandler_trap`.
- Interrupt handlers are divided into two stages:
    - `_start_INTERRUPT_trap`: saves space in stack, stores `a0` register, loads the interrupt handler in `a0`, and jumps to `_continue_interrupt_trap`.
    - `_continue_interrupt_trap`: saves the rest of the registers, calls the interrupt handler stored in `a0`, restores all registers, and returns.
- To help users to define interrupt traps together with their interrupt handler, the new `riscv_rt::interrupt` macro also generates the `_start_INTERRUPT_trap` routine.

romancardenas:
The PR is good to go. @rust-embedded/riscv please take a look.

riscv-rt/macros/src/lib.rs
```diff
@@ -312,3 +312,290 @@ pub fn loop_global_asm(input: TokenStream) -> TokenStream {
...
+.weak _start_trap
+_start_trap:
+    addi sp, sp, - {TRAP_SIZE} * {width}
+    {store}
+    add a0, sp, zero
+    jal ra, _start_trap_rust
+    {load}
+    addi sp, sp, {TRAP_SIZE} * {width}
+    {ret}
+\");"
```
rmsyn, line 442:
```suggestion
        r#"
core::arch::global_asm!(
".section .trap, \\"ax\\"
.align {width}
.weak _start_trap
_start_trap:
    addi sp, sp, - {TRAP_SIZE} * {width}
    {store}
    add a0, sp, zero
    jal ra, _start_trap_rust
    {load}
    addi sp, sp, {TRAP_SIZE} * {width}
    {ret}
");"#
```

riscv-rt/macros/src/lib.rs
```diff
@@ -312,3 +312,290 @@ pub fn loop_global_asm(input: TokenStream) -> TokenStream {
...
+    la a0, DefaultHandler               // load interrupt handler address into a0
+
+.global _continue_interrupt_trap
+_continue_interrupt_trap:
+    {store_continue}                   // store trap partially (all registers except a0)
+    jalr ra, a0, 0                     // jump to corresponding interrupt handler (address stored in a0)
+    {load}                             // restore trap frame
+    addi sp, sp, {TRAP_SIZE} * {width} // deallocate space for trap frame
+    {ret}                              // return from interrupt
+\");"
```
rmsyn, line 499:
```suggestion
        r#"
core::arch::global_asm!(
".section .trap, \\"ax\\"

.global _start_DefaultHandler_trap
_start_DefaultHandler_trap:
    addi sp, sp, -{TRAP_SIZE} * {width} // allocate space for trap frame
    {store_start}                       // store trap partially (only register a0)
    la a0, DefaultHandler               // load interrupt handler address into a0

.global _continue_interrupt_trap
_continue_interrupt_trap:
    {store_continue}                   // store trap partially (all registers except a0)
    jalr ra, a0, 0                     // jump to corresponding interrupt handler (address stored in a0)
    {load}                             // restore trap frame
    addi sp, sp, {TRAP_SIZE} * {width} // deallocate space for trap frame
    {ret}                              // return from interrupt
");"#
```

riscv-rt/macros/src/lib.rs
```diff
@@ -312,3 +312,290 @@ pub fn loop_global_asm(input: TokenStream) -> TokenStream {
...
+core::arch::global_asm!(
+    \".section .trap, \\\"ax\\\"
+    .align 2
+    .global _start_{interrupt}_trap
+    _start_{interrupt}_trap:
+        addi sp, sp, -{TRAP_SIZE} * {width} // allocate space for trap frame
+        {store}                             // store trap partially (only register a0)
+        la a0, {interrupt}                  // load interrupt handler address into a0
+        j _continue_interrupt_trap          // jump to common part of interrupt trap
+\");"
```
rmsyn, line 597:
```suggestion
        r#"
core::arch::global_asm!(
    ".section .trap, \\"ax\\"
    .align 2
    .global _start_{interrupt}_trap
    _start_{interrupt}_trap:
        addi sp, sp, -{TRAP_SIZE} * {width} // allocate space for trap frame
        {store}                             // store trap partially (only register a0)
        la a0, {interrupt}                  // load interrupt handler address into a0
        j _continue_interrupt_trap          // jump to common part of interrupt trap
");"#
```

riscv-rt/src/asm.rs
```diff
@@ -274,64 +277,46 @@ _pre_init_trap:
...
+        j _start_MachineTimer_trap
+        j _start_DefaultHandler_trap      // Interrupt 8 is reserved
+        j _start_SupervisorExternal_trap
+        j _start_DefaultHandler_trap      // Interrupt 10 is reserved
+        j _start_MachineExternal_trap
+
+        // default table does not include the remaining interrupts.
+        // Targets with extra interrupts should override this table.
+    
+    .option pop",
```
rmsyn, line 319:
```suggestion
    r#".section .trap, "ax"
    .weak _vector_table
    .type _vector_table, @function
    
    .option push
    .balign 0x4 // TODO check if this is the correct alignment
    .option norelax
    .option norvc
    
    _vector_table:
        j _start_trap                     // Interrupt 0 is used for exceptions
        j _start_SupervisorSoft_trap
        j _start_DefaultHandler_trap      // Interrupt 2 is reserved
        j _start_MachineSoft_trap
        j _start_DefaultHandler_trap      // Interrupt 4 is reserved
        j _start_SupervisorTimer_trap
        j _start_DefaultHandler_trap      // Interrupt 6 is reserved
        j _start_MachineTimer_trap
        j _start_DefaultHandler_trap      // Interrupt 8 is reserved
        j _start_SupervisorExternal_trap
        j _start_DefaultHandler_trap      // Interrupt 10 is reserved
        j _start_MachineExternal_trap

        // default table does not include the remaining interrupts.
        // Targets with extra interrupts should override this table.
    
    .option pop"#,
```

riscv-rt/src/asm.rs
```diff
@@ -274,64 +277,46 @@ _pre_init_trap:
...
+
+#[cfg(feature = "v-trap")]
+cfg_global_asm!(
+    // Set the vector mode to vectored.
+    ".section .trap, \"ax\"
+    .weak _vector_table
+    .type _vector_table, @function
+    
+    .option push
+    .balign 0x4 // TODO check if this is the correct alignment
```
rmsyn, line 298: This looks correct to me, from the RISC-V Privileged Spec entry for `mtvec`:

```
Asynchronous interrupts set pc to BASE+4×cause.
```

riscv-rt/macros/src/lib.rs
```diff
@@ -312,3 +312,290 @@ pub fn loop_global_asm(input: TokenStream) -> TokenStream {
...
+fn load_trap(arch: RiscvArch) -> String {
+    let (width, load) = match arch {
+        RiscvArch::Rv32 => (4, "lw"),
+        RiscvArch::Rv64 => (8, "ld"),
+    };
+    let mut loads = Vec::new();
+    for (i, reg) in TRAP_FRAME.iter().enumerate() {
+        loads.push(format!("{load} {reg}, {i}*{width}(sp)"));
+    }
+    loads.join("\n")
```
rmsyn, line 381:
```suggestion
    TRAP_FRAME.iter().enumerate().map(|i, reg| {
        let end = if i < TRAP_FRAME.len() - 1 { "\n" } else { "" };
        format!("{load} {reg}, {i}*{width}(sp){end}")
    }).collect()
```

Saves an extra allocation.

hegza: I'd use rmsyn's solution but with the original `join` operation included. To me, it's more readable than the `if` and I'd assume that performance isn't affected by that.

riscv-rt/src/asm.rs
hegza: The [CLIC](https://github.com/riscv/riscv-fast-interrupt/blob/master/clic.adoc) specification may require `_vector_table` to be 64-byte aligned so this will have to be overridden but I think there's not much we can or should do in `riscv-rt` about this.

romancardenas: Also, if I remember well, CLIC interrupt table is not comprised of `j` instructions, but the address of the trap itself. So, once we start integrating vectored CLIC mode, we might use a different methodology.

hegza: This is correct. CLIC uses a vector table of addresses instead of jumps (at least our implementation 🙂).

MabezDev:
Sorry for not taking a look at this sooner. I will hopefully have some time to review at the weekend :)

riscv-rt/macros/src/lib.rs
romancardenas: Done!

rmsyn (commented):
Changes LGTM. I'll work on testing these changes on hardware soon.

riscv-rt/macros/src/lib.rs
```diff
@@ -312,3 +312,290 @@ pub fn loop_global_asm(input: TokenStream) -> TokenStream {
...
+    let ret = "sret";
+    #[cfg(not(feature = "s-mode"))]
+    let ret = "mret";
+
+    format!(
+        r#"
+core::arch::global_asm!(
+".section .trap, \"ax\"
+.align {width}
+.weak _start_trap
```
MabezDev, line 433: One interesting data point, which might not be required here, is that in esp-riscv-rt, with high opt level + lto = 'fat' we found instances where weak assembly symbols were duplicated by the linker. To fix this we had to sprinkle some weird llvm specific attributes before the weak definitions:

https://github.com/esp-rs/esp-hal/blob/3c057595561a7a5dba68ae4df09c9960217c33fc/esp-riscv-rt/src/lib.rs#L830-L835

_probably_ not required here, but if you run into issue you'll know what to do :D.

MabezDev (approved):
This LGTM, I hope in the some day we can ditch esp-riscv-rt and go back to using riscv-rt and friends :)

riscv-rt/macros/src/lib.rs
romancardenas: Thank you! This is my goal. If `riscv-rt` is also suitable for ESP32 Cx, then it is flexible enough for any RISC-V target. The issue you point out is interesting, we will monitor it in case it appears in future versions

Issue #187, opened 2024-02-22
`riscv`: Pub macros for non-standard CSRs


dreiss:
I'm using Rust on a RISC-V processor that includes some non-standard CSRs.  It would be nice if the macros this crate uses to define CSR functions were exported so I could use them to define my own CSR functions.

romancardenas:
Sounds good to me. Feel free to open a PR :)

PR #203, opened 2024-05-01
`riscv`: register: exports macros for custom CSRs

rmsyn:
Exports macros from the `riscv::register::macros` module for external users.

This allows external users to define custom CSR accessor functions.

This is useful for platforms and SoCs with custom CSR registers not defined in the `riscv` crate.

Resolves: #187

### DIFF
--- a/riscv/src/register/macros.rs
+++ b/riscv/src/register/macros.rs
@@ -1,3 +1,9 @@
+/// `RV64`: Convenience macro to wrap the `csrrs` assembly instruction for reading a CSR register.
+///
+/// This macro should generally not be called directly.
+///
+/// Instead, use the [read_csr_as](crate::read_csr_as) or [read_csr_as_usize](crate::read_csr_as_usize) macros.
+#[macro_export]
 macro_rules! read_csr {
     ($csr_number:literal) => {
         /// Reads the CSR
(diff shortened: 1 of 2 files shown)


riscv/src/register/macros.rs
```diff
@@ -1,3 +1,9 @@
+/// `RV64`: Convenience macro to wrap the `csrrs` assembly instruction for reading a CSR register.
+///
+/// This macro should generally not be called directly.
```
rmsyn, line 3: Not sure if this is actually correct, as some of the existing `register` definitions directly call this macro.

Open to wording suggestions about correct usage here, and other similar locations.

riscv/src/register/macros.rs
```diff
@@ -18,6 +24,12 @@ macro_rules! read_csr {
     };
 }
 
+/// `RV32`: Convenience macro to wrap the `csrrs` assembly instruction for reading a CSR register.
+///
+/// This macro should generally not be called directly.
+///
+/// Instead, use the [read_csr_as_rv32] or [read_csr_as_usize_rv32] macros.
```
rmsyn, line 31: These and similar doc links are broken. Will be fixed with next set of changes.

romancardenas (approved):
LGTM

riscv/src/register/macros.rs
```diff
@@ -1,3 +1,9 @@
+/// `RV64`: Convenience macro to wrap the `csrrs` assembly instruction for reading a CSR register.
```
taiki-e, line 1: This macro is for both RV32 and RV64. (Other non-`*_rv32` macros are the same.)

PR #204, opened 2024-05-03
`riscv`: register: fix target architecture conditional compilation

rmsyn:
Uses a conditional compilation selector set by modern compiler versions for code gated behind a target architecture.

### DIFF
--- a/riscv/src/register/macros.rs
+++ b/riscv/src/register/macros.rs
@@ -1,4 +1,4 @@
-/// `RV64`: Convenience macro to wrap the `csrrs` assembly instruction for reading a CSR register.
+/// Convenience macro to wrap the `csrrs` assembly instruction for reading a CSR register.
 ///
 /// This macro should generally not be called directly.
 ///
@@ -10,14 +10,14 @@ macro_rules! read_csr {
         #[inline]
         unsafe fn _read() -> usize {
             match () {
-                #[cfg(riscv)]
+                #[cfg(any(target_arch = "riscv32", target_arch = "riscv64"))]
                 () => {
                     let r: usize;
                     core::arch::asm!(concat!("csrrs {0}, ", stringify!($csr_number), ", x0"), out(reg) r);
                     r
                 }
 
-                #[cfg(not(riscv))]
+                #[cfg(not(any(target_arch = "riscv32", target_arch = "riscv64")))]
                 () => unimplemented!(),
             }
         }
@@ -36,21 +36,21 @@ macro_rules! read_csr_rv32 {
         #[inline]
         unsafe fn _read() -> usize {
             match () {
-                #[cfg(riscv32)]
+                #[cfg(target_arch = "riscv32")]
                 () => {
                     let r: usize;
                     core::arch::asm!(concat!("csrrs {0}, ", stringify!($csr_number), ", x0"), out(reg) r);
                     r
                 }
 
-                #[cfg(not(riscv32))]
+                #[cfg(not(target_arch = "riscv32"))]
                 () => unimplemented!(),
             }
         }
     };
 }
 
-/// `RV64`: Convenience macro to read a CSR register value as a `register` type.
+/// Convenience macro to read a CSR register value as a `register` type.
 ///
 /// The `register` type must be a defined type in scope of the macro call.
 #[macro_export]
@@ -86,7 +86,7 @@ macro_rules! read_csr_as_rv32 {
     };
 }
 
-/// `RV64`: Convenience macro to read a CSR register value as a [`usize`].
+/// Convenience macro to read a CSR register value as a [`usize`].
 #[macro_export]
 macro_rules! read_csr_as_usize {
     ($csr_number:literal) => {
@@ -114,7 +114,7 @@ macro_rules! read_csr_as_usize_rv32 {
     };
 }
 
-/// `RV64`: Convenience macro to wrap the `csrrw` assembly instruction for writing to CSR registers.
+/// Convenience macro to wrap the `csrrw` assembly instruction for writing to CSR registers.
 ///
 /// This macro should generally not be called directly.
 ///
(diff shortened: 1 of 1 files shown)


rmsyn:
If this is approved, I can open another PR to use these selectors in the rest of the library, or add more commits to do it here.

romancardenas (approved):

romancardenas:
Regarding using `target_arch` instead of custom compilation flags, I don't know what to say. We would still need flags such as `riscve` for ISA extensions.

rmsyn:
> Regarding using target_arch instead of custom compilation flags, I don't know what to say. 

How are these custom compilation flags added currently?

I ran into this issue because of a crate I'm working on `panic`ing during boot. Took a while to trace the cause to this conditional compilation.

> We would still need flags such as riscve for ISA extensions.

I think this would work for what we would want for RISC-V extensions, though I'm not sure what is present in stable/nightly Rust: <https://doc.rust-lang.org/reference/conditional-compilation.html#target_feature>

rmsyn:
Added some fixes for the rest of the `riscv` crate. Let me know if you would like me to split these changes into a separate PR instead.

taiki-e:
> How are these custom compilation flags added currently?

These are set by the build script:
https://github.com/rust-embedded/riscv/blob/a9d3e33ff9d321bb0b19e260c66e1d5fa7a4221c/riscv/build.rs#L4-L12

However, `target.starts_with("riscv32")`/`target.starts_with("riscv64")` is not very robust and will not work if the target is a custom target with a name is something like `riscv-unknown-*`. (Referring to `CARGO_CFG_TARGET_ARCH` instead of `TARGET` is a robust way here)

romancardenas:
> Added some fixes for the rest of the `riscv` crate. Let me know if you would like me to split these changes into a separate PR instead.

Please, do so in a separate PR. I am not sure whether it is worth it or not. Notice that feature gates become way more verbose. However, changing the build script as suggested by @taiki-e sounds more than reasonable.

rmsyn:
> However, changing the build script as suggested by @taiki-e sounds more than reasonable.

I was exploring this a bit, as well. I'll try the suggestion by @taiki-e. If it works, I'll apply that patch instead.

@taiki-e would you like to make the PR, or would you like me to?

rmsyn:
> I am not sure whether it is worth it or not. Notice that feature gates become way more verbose.

So, I A/B tested the changes in this PR and #205. The changes here work as expected, calls into the various `asm` and `register::macro` calls call the expected RISC-V assembly instructions.

With #205, the code `panic`s from the `unimplemented` functions. For some reason, the custom `cfg` arguments are not making it through to downstream users. Any thoughts?

I agree the verbosity is a bit of a problem, but a small price to pay for functioning code.

This is tested on bare-metal VisionFive2 board.

taiki-e:
> With #205, the code `panic`s from the `unimplemented` functions. For some reason, the custom `cfg` arguments are not making it through to downstream users. Any thoughts?

The cfg set by the build script affects only the crate where the build script is located.

And I think that is a bug that was missed in https://github.com/rust-embedded/riscv/pull/203. Exported macros need to reference `cfg(target_arch = "..")`.

riscv/src/register/macros.rs
```diff
@@ -10,14 +10,14 @@ macro_rules! read_csr {
         #[inline]
         unsafe fn _read() -> usize {
             match () {
-                #[cfg(riscv)]
+                #[cfg(target_arch = "riscv64")]
```
taiki-e, line 13: I think replacing `riscv` -> `target_arch = "riscv64"` in this file is not correct: https://github.com/rust-embedded/riscv/pull/203#discussion_r1590188559

rmsyn:
> Exported macros need to reference cfg(target_arch = "..").

So, maybe just the exported macros get `cfg(target_arch = "..")`, and the rest can use the existing `cfg(riscv*)`?

If so, I can drop the rest of the commits for code outside the `riscv::register::macros` module.

riscv/src/register/macros.rs
rmsyn: This was my mistake, I'll fix it in the next set of fixups to this PR. Along with the comments about non-`_rv32` macros.

riscv/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - Fixed `sip::set_ssoft` and `sip::clear_ssoft` using wrong address
+- Fixed conditional compilation selectors for `riscv::register::macros`
```
taiki-e, line 20: https://github.com/rust-embedded/riscv/pull/203 has not been released yet (i.e., users of the released version have never been affected by the bug), so I don't think the changelog entry for this is needed.

rmsyn:
I don't know why the `clippy` checks are suddenly failing now. I run the checks locally from CI runner:

```bash
cargo clippy --package riscv-rt --all --features=s-mode -- -D warnings
```

The checks pass. :thinking:

riscv/CHANGELOG.md
rmsyn: Where should it go then, or are you saying it doesn't need a `CHANGELOG.md` entry at all?

Asking because the CI check fails without it an entry.

romancardenas: I added a tag to skip the changelog CI

romancardenas (approved):
Looks good to me! I am OK with using `riscv`, `riscv32`, etc. in the crate but using `cfg(target_arch = _)` in public macros.

romancardenas:
Let's merge #205 before. It should fix the current clippy issues.

riscv/CHANGELOG.md
rmsyn: > I added a tag to skip the changelog CI

Sounds good, I'll remove it when I rebase the changes on #205.

romancardenas (approved):
LGTM

PR #205, opened 2024-05-04
riscv: build: make `cfg` variables more robust

rmsyn:
Change which Cargo `CFG` environment variables are used to select the custom `cfg` variables.

See https://github.com/rust-embedded/riscv/pull/204#issuecomment-2094089195 for discussion.

### DIFF
--- a/riscv-rt/build.rs
+++ b/riscv-rt/build.rs
@@ -72,6 +72,16 @@ fn parse_target(target: &str, cargo_flags: &str) -> (u32, HashSet<char>) {
 }
 
 fn main() {
+    println!("cargo:rustc-check-cfg=cfg(riscv)");
+    println!("cargo:rustc-check-cfg=cfg(riscv32)");
+    println!("cargo:rustc-check-cfg=cfg(riscv64)");
+    println!("cargo:rustc-check-cfg=cfg(riscvi)");
+    println!("cargo:rustc-check-cfg=cfg(riscvm)");
+    println!("cargo:rustc-check-cfg=cfg(riscva)");
+    println!("cargo:rustc-check-cfg=cfg(riscvf)");
+    println!("cargo:rustc-check-cfg=cfg(riscvd)");
+    println!("cargo:rustc-check-cfg=cfg(riscvc)");
+
     let target = env::var("TARGET").unwrap();
     let cargo_flags = env::var("CARGO_ENCODED_RUSTFLAGS").unwrap();
     let _name = env::var("CARGO_PKG_NAME").unwrap();
@@ -79,7 +89,6 @@ fn main() {
     // set configuration flags depending on the target
     if target.starts_with("riscv") {
         println!("cargo:rustc-cfg=riscv");
-
         // This is required until target_arch & target_feature risc-v work is
         // stable and in-use (rust 1.75.0)
         let (bits, extensions) = parse_target(&target, &cargo_flags);
--- a/riscv/CHANGELOG.md
+++ b/riscv/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 - Made `asm::wfi`, `fence`, `fence_i` and `sfence` safe (ie, removed `unsafe` from their definitions)
+- Made `cfg` variable selection more robust for custom targets
 
 ## [v0.11.0] - 2024-01-14
 
--- a/riscv/build.rs
+++ b/riscv/build.rs
@@ -1,6 +1,10 @@
 use std::env;
 
 fn main() {
+    println!("cargo:rustc-check-cfg=cfg(riscv)");
+    println!("cargo:rustc-check-cfg=cfg(riscv32)");
+    println!("cargo:rustc-check-cfg=cfg(riscv64)");
+
     let target_arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap();
 
     if target_arch == "riscv32" {
--- a/riscv/build.rs
+++ b/riscv/build.rs
@@ -1,12 +1,12 @@
 use std::env;
 
 fn main() {
-    let target = env::var("TARGET").unwrap();
+    let target_arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap();
 
-    if target.starts_with("riscv32") {
+    if target_arch == "riscv32" {
         println!("cargo:rustc-cfg=riscv");
         println!("cargo:rustc-cfg=riscv32");
-    } else if target.starts_with("riscv64") {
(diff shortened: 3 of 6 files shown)


riscv/build.rs
```diff
@@ -1,12 +1,13 @@
...
-    let target = env::var("TARGET").unwrap();
+    let target = env::var("CARGO_CFG_TARGET_ARCH").unwrap();
+    let ptr_width = env::var("CARGO_CFG_TARGET_POINTER_WIDTH").unwrap();
 
-    if target.starts_with("riscv32") {
+    if target.contains("riscv") && ptr_width == "32" {
         println!("cargo:rustc-cfg=riscv");
         println!("cargo:rustc-cfg=riscv32");
-    } else if target.starts_with("riscv64") {
+    } else if target.contains("riscv") && ptr_width == "64" {
```
taiki-e, line 10: It would be better to compare a full `target_arch` instead of referring to a `target_pointer_width`. (It is more simple and there is also a [proposal on 32-bit (pointer) ABI of riscv64](https://lwn.net/Articles/951187/).)

```rs
    let target_arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap();

    if target_arch == "riscv32" {
        println!("cargo:rustc-cfg=riscv");
        println!("cargo:rustc-cfg=riscv32");
    } else if target_arch == "riscv64" {
```

riscv/CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 - Made `asm::wfi`, `fence`, `fence_i` and `sfence` safe (ie, removed `unsafe` from their definitions)
+- Made `cfg` variable selection more robust for custom targets and downstream users
```
taiki-e, line 26:
```suggestion
- Made `cfg` variable selection more robust for custom targets
```

rmsyn:
I don't know why the `clippy` checks are suddenly failing now. I run the checks locally from CI runner:

```bash
cargo clippy --package riscv-rt --all --features=s-mode -- -D warnings
```

The checks pass. :thinking:

romancardenas:
In nightly, clippy now complains about "custom" compiler features (check [this](https://doc.rust-lang.org/nightly/cargo/reference/build-scripts.html#rustc-check-cfg)).

So, as suggested by the CI, now we should add to `build.rs` something like:
```rust
println!("cargo::rustc-check-cfg=cfg(riscv)");
println!("cargo::rustc-check-cfg=cfg(riscv32)");
println!("cargo::rustc-check-cfg=cfg(riscv64)");
```

Also, as we need to modify `build.rs`, it might be a good moment to address @taiki-e 's comments [here](https://github.com/rust-embedded/riscv/pull/205#discussion_r1590187609).

romancardenas (changes requested):
Looks good to me! However, we still need to address new clippy requirements

rmsyn:
Build checks on `1.60` machines are failing because of bumped MSRV for the `cfg` checks.

Without an MSRV bump, the `clippy` checks have the following error:

```bash
cargo +nightly clippy --all --no-default-features --target riscv64imac-unknown-none-elf -- -D warnings
   Compiling riscv v0.11.1 (/riscv/riscv)
   Compiling riscv-rt v0.13.0 (/riscv/riscv-rt)
error: the `cargo::` syntax for build script output instructions was added in Rust 1.77.0, but the minimum supported Rust version of `riscv v0.11.1 (/riscv/riscv)` is 1.60.
See https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script for more information about build script outputs.
```

riscv/build.rs
```diff
@@ -1,12 +1,16 @@
 use std::env;
 
 fn main() {
-    let target = env::var("TARGET").unwrap();
+    println!("cargo::rustc-check-cfg=cfg(riscv)");
+    println!("cargo::rustc-check-cfg=cfg(riscv32)");
```
rmsyn, line 5: We might want to move these checks (`riscv32` + `riscv64`) into their respective conditional branches.

All the checks pass, so I'm not 100% on how these checks are supposed to work.

romancardenas: with `cargo:rustc-check-cfg` you just inform cargo that you might have a custom `cfg` item.

romancardenas:
I don't like the idea of bumping MSRV from 1.60 to 1.77 due to a nightly clippy issue... I looked at the line provided by your post and:

> Note: The old invocation prefix cargo: (one colon only) is deprecated and won’t get any new features. To migrate, use two-colons prefix cargo::, which was added in Rust 1.77. If you were using cargo:KEY=VALUE for arbitrary links manifest key-value pairs, it is encouraged to switch to cargo::metadata=KEY=VALUE. Stick to cargo: only if the support of Rust version older than 1.77 is required.

I suggest using just `:` instead of `::` until we face this issue in stable clippy.

riscv/build.rs
```diff
@@ -1,12 +1,16 @@
...
+    println!("cargo::rustc-check-cfg=cfg(riscv64)");
```
romancardenas, line 6:
```suggestion
    println!("cargo:rustc-check-cfg=cfg(riscv)");
    println!("cargo:rustc-check-cfg=cfg(riscv32)");
    println!("cargo:rustc-check-cfg=cfg(riscv64)");
```

romancardenas: Using only  single colon should do the trick and avoid bumping the MSRV

romancardenas (changes requested):
Let's see if we can get rid of bumping the MSRV

riscv-rt/build.rs
```diff
@@ -100,6 +104,7 @@ fn main() {
 
         // expose the ISA extensions
         for ext in &extensions {
+            println!("cargo::rustc-check-cfg=cfg(riscv{ext})");
```
jsgf, line 107: This is not enough - it will only add the extensions of the current target, not all the extensions we may be checking for.

rmsyn: Right, this lint check is new to me, and I thought it was for checking expected `cfg` output.

After re-reading the docs, it's clear now that it's meant to capture **all** expected `cfg` output.

I'll fix it in the next set of changes.

rmsyn:
> I don't like the idea of bumping MSRV from 1.60 to 1.77 due to a nightly clippy issue

I didn't like it either. Looks like the single colon fixed the warning/error. Weird that the compiler would suggest the non-MSRV compatible fix... Probably pretty involved for it to do otherwise.

Either way, the CI looks like it's fixed now. Thank you for the suggestion @romancardenas.

romancardenas (approved):
LGTM

PR #206, opened 2024-05-07
Implement Eq and PartialEq for Range and Permission

jsgf:
It's useful to be able to do equality tests for these for unit tests.

Should we do this more consistently?


romancardenas (changes requested):
I like this approach! And also I think it would be useful to add it to other structures. Please, fix the conflicts before merging this to master.

jsgf:
OK I rebased it over the check-cfg updates.

PR #207, opened 2024-05-07
Add Mstatus helpers to allow setting fields in Mstatus

jsgf:
Without needing to touch the CSR. This allows multiple changes in a single register write.

### DIFF
--- a/riscv/src/lib.rs
+++ b/riscv/src/lib.rs
@@ -36,6 +36,7 @@
 #![allow(clippy::missing_safety_doc)]
 
 pub mod asm;
+pub(crate) mod bits;
 pub mod delay;
 pub mod interrupt;
 pub mod register;
--- a/riscv/src/register/macros.rs
+++ b/riscv/src/register/macros.rs
@@ -160,6 +160,34 @@ macro_rules! write_csr_rv32 {
     };
 }
 
+/// Convenience macro to write a value with `bits` to a CSR
+#[macro_export]
+macro_rules! write_csr_as {
+    ($csr_type:ty, $csr_number:literal) => {
+        $crate::write_csr!($csr_number);
+
+        /// Writes the CSR
+        #[inline]
+        pub fn write(value: $csr_type) {
+            unsafe { _write(value.bits) }
+        }
+    };
+}
+
+/// Convenience macro to write a value to a CSR register.
+#[macro_export]
+macro_rules! write_csr_as_rv32 {
+    ($csr_type:ty, $csr_number:literal) => {
+        $crate::write_csr_rv32!($csr_number);
+
+        /// Writes the CSR
+        #[inline]
+        pub fn write(value: $csr_type) {
+            unsafe { _write(value.bits) }
+        }
+    };
+}
+
 /// Convenience macro to write a [`usize`] value to a CSR register.
 #[macro_export]
 macro_rules! write_csr_as_usize {
--- a/riscv/src/register/mstatus.rs
+++ b/riscv/src/register/mstatus.rs
@@ -481,7 +481,7 @@ impl Mstatus {
 }
 
 read_csr_as!(Mstatus, 0x300);
-write_csr!(0x300);
+write_csr_as!(Mstatus, 0x300);
 set!(0x300);
 clear!(0x300);
 
--- a/riscv/src/register/mstatus.rs
+++ b/riscv/src/register/mstatus.rs
@@ -1,6 +1,7 @@
 //! mstatus register
 
 pub use super::misa::XLEN;
+use crate::bits::{bf_extract, bf_insert};
 
 /// mstatus register
 #[derive(Clone, Copy, Debug)]
@@ -75,10 +76,15 @@ impl Mstatus {
     /// Helper to insert a bitfield into Mstatus
     #[inline]
     fn bf_insert(&self, bit: usize, width: usize, val: usize) -> Self {
-        let mask = (1 << width) - 1;
         Self {
-            bits: self.bits & !(mask << bit) | ((val & mask) << bit),
+            bits: bf_insert(self.bits, bit, width, val),
         }
+    }
+
+    /// Helper to extract a bitfield from Mstatus
+    #[inline]
+    fn bf_extract(&self, bit: usize, width: usize) -> usize {
+        bf_extract(self.bits, bit, width)
     }
 
     /// Returns the contents of the register as raw bits
@@ -90,7 +96,7 @@ impl Mstatus {
     /// Supervisor Interrupt Enable
     #[inline]
     pub fn sie(&self) -> bool {
-        self.bits & (1 << 1) != 0
+        self.bf_extract(1, 1) != 0
     }
 
     /// Update Supervisor Interrupt Enable
@@ -106,7 +112,7 @@ impl Mstatus {
     /// Machine Interrupt Enable
     #[inline]
     pub fn mie(&self) -> bool {
-        self.bits & (1 << 3) != 0
+        self.bf_extract(3, 1) != 0
     }
 
     /// Update Machine Interrupt Enable
@@ -122,7 +128,7 @@ impl Mstatus {
     /// Supervisor Previous Interrupt Enable
     #[inline]
     pub fn spie(&self) -> bool {
-        self.bits & (1 << 5) != 0
+        self.bf_extract(5, 1) != 0
     }
 
     /// Update Supervisor Previous Interrupt Enable
@@ -138,7 +144,7 @@ impl Mstatus {
     /// U-mode non-instruction-fetch memory endianness
     #[inline]
     pub fn ube(&self) -> Endianness {
-        Endianness::from(self.bits & (1 << 6) != 0)
+        Endianness::from(self.bf_extract(6, 1) != 0)
     }
 
     /// Update U-mode non-instruction-fetch memory endianness
@@ -154,7 +160,7 @@ impl Mstatus {
     /// Machine Previous Interrupt Enable
     #[inline]
     pub fn mpie(&self) -> bool {
-        self.bits & (1 << 7) != 0
+        self.bf_extract(7, 1) != 0
     }
 
     /// Update Machine Previous Interrupt Enable
@@ -170,7 +176,7 @@ impl Mstatus {
     /// Supervisor Previous Privilege Mode
     #[inline]
     pub fn spp(&self) -> SPP {
-        match self.bits & (1 << 8) != 0 {
+        match self.bf_extract(7, 1) != 0 {
             true => SPP::Supervisor,
             false => SPP::User,
         }
@@ -189,7 +195,7 @@ impl Mstatus {
     /// Machine Previous Privilege Mode
     #[inline]
     pub fn mpp(&self) -> MPP {
-        let mpp = (self.bits >> 11) & 0x3; // bits 11-12
+        let mpp = self.bf_extract(11, 2); // bits 11-12
         match mpp {
             0b00 => MPP::User,
             0b01 => MPP::Supervisor,
@@ -214,7 +220,7 @@ impl Mstatus {
     /// and floating-point data registers `f0–f31`.
     #[inline]
     pub fn fs(&self) -> FS {
-        let fs = (self.bits >> 13) & 0x3; // bits 13-14
+        let fs = self.bf_extract(13, 2); // bits 13-14
         match fs {
             0b00 => FS::Off,
             0b01 => FS::Initial,
@@ -240,7 +246,7 @@ impl Mstatus {
     /// state.
     #[inline]
     pub fn xs(&self) -> XS {
-        let xs = (self.bits >> 15) & 0x3; // bits 15-16
+        let xs = self.bf_extract(15, 2); // bits 15-16
         match xs {
             0b00 => XS::AllOff,
             0b01 => XS::NoneDirtyOrClean,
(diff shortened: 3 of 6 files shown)


romancardenas:
I find it confusing having methods that modify `MStatus` that do not modify the mstatus register, but a "copy". Alternatively, I like the idea of register builders (see #98). However, the PR looks stale...

Would you consider adapting the builders approach for this PR?

jsgf:
I think there's a few things to pick apart here:
- a type representing possible values of mstatus
- the actual mstatus csr
- updating the csr with a new value

The third one is what adds the complications, because of the special atomic bit set/clear operations in a single csr instruction. These are very useful for all the single-bit fields in mstatus so we wouldn't want to do without them. But in the general case - either updating a multi-bit field or updating multiple fields at once - requires a full read/modify/write.

So as I see it, the current "operate directly on mstatus" methods are the special case exception, and the common case is:
```
let mut mstatus = mstatus::read();
mstatus = mstatus.set_X().set_Y();
mstatus::write(mstatus);
```

> Would you consider adapting the builders approach for this PR?

I'm not sure how that would really be different? As I said, my use-case here is to read mstatus, make some changes, then put it back, whereas a builder would imply building new Mstatus values from scratch.

But I agree with you about the potential for confusion. I should definitely add a clarifying line in the doc comment to say it updates the Mstatus value but not the csr, and maybe rename them to `update_`?

jsgf:
I added a commit to:
- add clarifying comment
- rename the methods from set_ -> update_ to make them more distinct from the `mstatus::set_` functions.

jsgf:
Oh I just noticed there's no `write_as` for updating a csr from a type. And mstatus doesn't have write_as_usize...

jsgf:
Hm, need to do more testing.

jsgf:
Ugh, shift vs `&` precedence. Looks like this is correct now.

riscv/src/register/mstatus.rs
```diff
@@ -84,30 +93,75 @@ impl Mstatus {
         self.bits & (1 << 1) != 0
     }
 
+    /// Update Supervisor Interrupt Enable
+    ///
+    /// Note this updates the [`Mstatus`] value, but does not affect the mstatus
+    /// CSR itself. See [`set_sie`]/[`clear_sie`] to directly update the CSR.
+    #[inline]
+    pub fn update_sie(&self, sie: bool) -> Self {
```
rmsyn, line 101: Somewhat of a bikeshed, but methods that set a bit in a bitfield without clearing the rest of the bitfield typically use the `modify_*` convention, e.g. in `pac` crates generated with `svd2rust`.

romancardenas:
Whenever I have doubts about new features, I take a look at how `cortex-m` works. For instance, the [FPSCR](https://github.com/rust-embedded/cortex-m/blob/master/cortex-m/src/register/fpscr.rs). They do have `x` methods to get bit fields and `set_x` methods to set bit fields in the read value. Next, they provide a `write` function that takes the register structure and writes its value in the register using inline assembly. This looks very similar to your proposal @jsgf 

Let me take some time to review your changes, I'll try to provide a review ASAP.

riscv/src/register/macros.rs
```diff
@@ -160,6 +160,34 @@ macro_rules! write_csr_rv32 {
...
+        #[inline]
+        pub fn write(value: $csr_type) {
+            unsafe { _write(value.bits) }
+        }
+    };
+}
+
 /// `RV64`: Convenience macro to write a [`usize`] value to a CSR register.
 #[macro_export]
 macro_rules! write_csr_as_usize {
```
romancardenas, line 193: This could be substituted by `write_csr_as(usize, csr_number)`, right?

riscv/src/register/mstatus.rs
```diff
@@ -72,6 +72,15 @@ impl From<bool> for Endianness {
 }
 
 impl Mstatus {
```
romancardenas, line 74: It would be nice to have macros for structs like `Mstatus`, as I expect a lot of duplicates with this

riscv/src/register/mstatus.rs
romancardenas: In `cortex-m`, they use the `set_*` convention. I would align with them

riscv/src/register/mstatus.rs
```diff
@@ -84,30 +93,75 @@ impl Mstatus {
         self.bits & (1 << 1) != 0
     }
 
+    /// Update Supervisor Interrupt Enable
+    ///
+    /// Note this updates the [`Mstatus`] value, but does not affect the mstatus
+    /// CSR itself. See [`set_sie`]/[`clear_sie`] to directly update the CSR.
+    #[inline]
+    pub fn update_sie(&self, sie: bool) -> Self {
+        self.bf_insert(1, 1, sie as usize)
```
romancardenas, line 102: perhaps using a `bit_set(pos: usize, val:bool`? it would be slightly faster when width is 1

riscv/src/register/mstatus.rs
```diff
@@ -84,30 +93,75 @@ impl Mstatus {
...
+
     /// Machine Interrupt Enable
     #[inline]
     pub fn mie(&self) -> bool {
         self.bits & (1 << 3) != 0
     }
 
+    /// Update Machine Interrupt Enable
+    ///
+    /// Note this updates the [`Mstatus`] value, but does not affect the mstatus
```
romancardenas, line 113:
```suggestion
    /// Note this updates the [`Mstatus`] value read previously, but does not affect the mstatus
```

riscv/src/register/mstatus.rs
romancardenas: Maybe we could have a `pub(crate) bit` module somewhere with bitwise **functions** for `usize`. In this way, we avoid having methods in every register, and all registers use the same function. I'm not sure if this would help us optimize the binary size or if the compiler is smart enough to collapse all these methods (don't think so).

romancardenas (changes requested):
Let's follow the `set_*` approach as in `cortex-m`. Also, maybe it is a good moment to add a utility module with bit-wise functions that can be shared among all the registers (it feels like we will need it in the future)

riscv/src/register/mstatus.rs
jsgf: I started with set but the concern was it would be confusing vs the functions which act directly on the csr. Also there's a semantic difference as the csr functions use parameterleas set/clear for single bit fields whereas the update/modify/... on Mstatus takes a parameter.

riscv/src/register/macros.rs
jsgf: No, because it assumes a `bits` field.

riscv/src/register/mstatus.rs
jsgf: I'd be surprised if it made any difference - `bf_insert` is inlined so the compiler can see the constants and optimize accordingly.

riscv/src/register/mstatus.rs
jsgf: Yeah I saw a lot of opportunity to collapse some repeated code.

jsgf:
@romancardenas Yeah, I think a good general structure for any register is to define a type to abstract all the bitfields, read/write operations in terms of that type, and operations on the type to get/set each field. And as an optimization, specific atomic set operations directly on the register if it makes a difference (as it does for riscv single bit fields, and multi-bit in some situations). But it's not really worth having those direct-access functions for anything that's going to require a full read/modify/write pattern anyway.

`mstatus` is just where I've run into this for my current project, but this pattern applies to pretty much everything in riscv::registers.

For registers which only contain one logical field - eg the pmp_addr registers - you could say that it's probably not necessary to define a new type for it. But that 2 bit shift keeps biting me, so I'd probably still go for it...

riscv/src/lib.rs
```diff
@@ -36,6 +36,7 @@
 #![allow(clippy::missing_safety_doc)]
 
 pub mod asm;
+pub(crate) mod bits;
```
romancardenas, line 39: we can leave it as this, but now I don't find any reason why we should hide it from dependents. Maybe PACs find them handy

riscv/src/register/mstatus.rs
```diff
@@ -72,6 +73,20 @@ impl From<bool> for Endianness {
...
+            bits: bf_insert(self.bits, bit, width, val),
+        }
+    }
+
+    /// Helper to extract a bitfield from Mstatus
+    #[inline]
+    fn bf_extract(&self, bit: usize, width: usize) -> usize {
+        bf_extract(self.bits, bit, width)
+    }
+
```
romancardenas, line 89:
```suggestion
```

riscv/src/register/mstatus.rs
```diff
@@ -81,46 +96,106 @@ impl Mstatus {
     /// Supervisor Interrupt Enable
     #[inline]
     pub fn sie(&self) -> bool {
-        self.bits & (1 << 1) != 0
+        self.bf_extract(1, 1) != 0
```
romancardenas, line 99:
```suggestion
        bf_extract(self.bits, 1, 1) != 0
```

riscv/src/register/mstatus.rs
```diff
@@ -81,46 +96,106 @@ impl Mstatus {
...
+    }
+
+    /// Update Supervisor Interrupt Enable
+    ///
+    /// Note this updates a previously read [`Mstatus`] value, but does not
+    /// affect the mstatus CSR itself. See [`set_sie`]/[`clear_sie`] to directly
+    /// update the CSR.
+    #[inline]
+    pub fn set_sie(&self, sie: bool) -> Self {
+        self.bf_insert(1, 1, sie as usize)
```
romancardenas, line 109:
```suggestion
        self.bits = bf_insert(self.bits, 1, 1, sie as usize);
```

riscv/src/register/mstatus.rs
```diff
@@ -81,46 +96,106 @@ impl Mstatus {
...
+        self.bf_extract(1, 1) != 0
+    }
+
+    /// Update Supervisor Interrupt Enable
+    ///
+    /// Note this updates a previously read [`Mstatus`] value, but does not
+    /// affect the mstatus CSR itself. See [`set_sie`]/[`clear_sie`] to directly
+    /// update the CSR.
+    #[inline]
+    pub fn set_sie(&self, sie: bool) -> Self {
```
romancardenas, line 108:
```suggestion
    pub fn set_sie(&mut self, sie: bool) {
```

riscv/src/register/mstatus.rs
romancardenas: I personally prefer using the new utility functions instead of filling all the registers with new methods. But it is not a hard feeling, if you think this approach has advantages, please write them so we can choose the best option.

riscv/src/register/mstatus.rs
romancardenas: From [`cortex-m`](https://github.com/rust-embedded/cortex-m/blob/4395baee03d57a4e1763c427073da5f33ec3b17e/cortex-m/src/register/fpscr.rs#L32):
```rust
#[inline]
pub fn set_n(&mut self, n: bool) {
    let mask = 1 << 31;
    match n {
        true => self.bits |= mask,
        false => self.bits &= !mask,
    }
}
```
So let's stick with the classic setter approach with mutable reference and no return value.

romancardenas (changes requested):
Thanks for the hard work! Have a look to make comments, especially the one regarding the setter approach.

riscv/src/register/mstatus.rs
jsgf: What about returning `&mut Self` so you can chain them?

riscv/src/lib.rs
jsgf: I'd hold off on that until we're confident it's a stable public api.

riscv/src/register/mstatus.rs
jsgf: It's mostly to save on repeating `Self { bits: ... }` but if we're going with the `&mut self` approach there's no benefit.

jsgf:
BTW there's a lot of repetition here and in other modules - what do you think about a larger scale macro to generate all of this more automatically? It would make it easier to have all the different registers be consistent in form.

(I'm not a huge fan of big macros, but this crate already has a fair amount of macro-driven codegen and I think it tips the cost/benefit scales the right way for this.)

riscv/src/register/mstatus.rs
```diff
@@ -81,46 +82,106 @@ impl Mstatus {
...
+    #[inline]
+    pub fn set_mpie(&mut self, mpie: bool) {
+        self.bits = bf_insert(self.bits, 7, 1, mpie as usize);
     }
 
     /// Supervisor Previous Privilege Mode
     #[inline]
     pub fn spp(&self) -> SPP {
-        match self.bits & (1 << 8) != 0 {
+        match bf_extract(self.bits, 7, 1) != 0 {
```
romancardenas, line 165:
```suggestion
        match bf_extract(self.bits, 8, 1) != 0 {
```

romancardenas (changes requested):
found a 🐛

riscv/src/register/mstatus.rs
romancardenas: I like your idea!

romancardenas:
I agree with you. We could explore using macros to "ease" the new changes to all the registers

riscv/src/register/mstatus.rs
jsgf: Ugh oops. Definitely need a macro-based mechanism that can derive all this from a single source of truth.

romancardenas (approved):
LGTM

PR #208, opened 2024-05-07
Fix nightly CI by adding --check-cfg flags

jsgf:
Emit `cargo::rustc-check-cfg=CHECK_CFG` in build.rs so that they can be cross-checked even if cfg option itself isn't set.


jsgf:
This is a subset og #205, aside from (I think) handling the `riscv{ext}` cfg flags better.

riscv-rt/build.rs
```diff
@@ -76,6 +76,13 @@ fn main() {
     let cargo_flags = env::var("CARGO_ENCODED_RUSTFLAGS").unwrap();
     let _name = env::var("CARGO_PKG_NAME").unwrap();
 
+    println!("cargo:rustc-check-cfg=cfg(riscv)");
+    println!("cargo:rustc-check-cfg=cfg(riscv32)");
+    println!("cargo:rustc-check-cfg=cfg(riscv64)");
+    for ext in ['i', 'e', 'm', 'a', 'f', 'd', 'g', 'c'] {
```
romancardenas, line 82: I like this approach. Could you open a new PR with this modification, please? I think the rest of the crates are already fixed.

romancardenas (changes requested):

jsgf:
Ok rebased and just reduced to using a loop for per-ext checks.

riscv-rt/build.rs
jsgf: Oh I just recycled this PR.

romancardenas (approved):
LGTM

PR #211, opened 2024-05-17
Support "non-standard" interrupts and exceptions

romancardenas:
Currently, the `riscv` and `riscv-rt` crates assume that the interrupt and exception sources are the ones gathered in the [RISCV ISA](https://drive.google.com/file/d/17GeetSnT5wW3xNuAHI95-SI1gPGd5sJ_/view?usp=drive_link). However, we do not support target-specific sources. Moreover, some RISC-V targets do not follow the standard interrupt scheme (e.g., ESP32 C3). Another example is the E310x microcontroller, which does not support supervisor interrupt sources.

# Proposal

Now that we have the `riscv-pac` trait to define interrupt and exception sources,  we should be able to adapt the rest of the crates to rely on generic functions with constraints to these traits. For example, the `mcause` register can have something like this:

```rust
// in riscv::register::mcause 
pub use riscv_pac::{CoreInterruptNumber, ExceptionNumber};

pub enum Trap<I, E> {
    Interrupt(I),
    Exception(E),
}

impl Mcause {
    pub fn cause<I: CoreInterruptNumber, E: ExceptionNumber>(&self) -> Trap<I, E> {
        if self.is_interrupt() {
            Trap::Interrupt(I::from_number(self.code()).unwrap())
        } else {
            Trap::Exception(E::from_number(self.code()).unwrap())
        }
    }
}
```

By default, `I` and `E` should be the standard enumerations (as we are doing now). Still, PACs will be able to opt out of this default implementation and provide their custom interrupt/exception enumeration.

Next, we can enrich `riscv-rt` macros to automatically generate the necessary code for letting PACs inject their custom interrupt and exception tables. Also, in the case of vectored mode, we could generate the necessary vector table and so on.

**NOTE THAT THIS IS JUST MY PROPOSAL**, but I would love to hear your opinion as well as clever mechanisms to implement this neatly.

I will do a self-review to point out some to-dos that I would like to discuss with you.

Solves #146


riscv-pac/src/lib.rs
```diff
@@ -1,15 +1,44 @@
...
+/// interrupts for a specific device. Alternatively, the `riscv` crate provides a default
+/// implementation for the RISC-V ISA. Each variant must convert to a `u16` of its interrupt number.
+///
+/// # Safety
+///
+/// * This trait must only be implemented on the `riscv` crate or on a PAC of a RISC-V target.
+/// * This trait must only be implemented on enums of interrupts.
 /// * Each enum variant must represent a distinct value (no duplicates are permitted),
 /// * Each enum variant must always return the same value (do not change at runtime).
 /// * All the interrupt numbers must be less than or equal to `MAX_INTERRUPT_NUMBER`.
```
romancardenas, line 44: Same with `InterruptNumber`. Should it be `usize`?

riscv-pac/src/lib.rs
```diff
@@ -26,11 +55,34 @@ pub unsafe trait InterruptNumber: Copy {
...
+///
+/// Core interrupts are interrupts are retrieved from the `mcause` CSR. Usually, vectored mode is
+/// only available for core interrupts. The `riscv` crate provides a default implementation for
+/// the RISC-V ISA. However, a PAC may override the default implementation if the target has a
+/// different interrupt numbering scheme (e.g., ESP32C3).
+///
+/// # Safety
+///
+/// Each enum variant must represent a valid core interrupt number read from the `mcause` CSR.
+pub unsafe trait CoreInterruptNumber: InterruptNumber {}
```
romancardenas, line 68: For interrupt codes that **can be read from xcause**

riscv-pac/src/lib.rs
```diff
@@ -26,11 +55,34 @@ pub unsafe trait InterruptNumber: Copy {
...
+///
+/// External interrupts are interrupts caused by external sources (e.g., GPIO, UART, SPI).
+/// External interrupts are **not** retrieved from the `mcause` CSR.
+/// Instead, RISC-V processors have a single core interrupt for all external interrupts.
+/// An additional peripheral (e.g., PLIC) is used to multiplex the external interrupts.
+///
+/// # Safety
+///
+/// Each enum variant must represent a valid external interrupt number.
+pub unsafe trait ExternalInterruptNumber: InterruptNumber {}
```
romancardenas, line 80: For interrupt codes handled by an external interrupt peripheral (e.g., CLIC or PLIC)

almindor:
I think this is a good approach once it's clean and tested. I only had time to skim through it but don't see anything "wrong" with it.

riscv-pac/src/lib.rs
rmsyn: Does this need to be an `unsafe` trait? `mcause` is a `WARL` register, so even if a user attempts to write an invalid value, no soundness guarantees will be violated.

Are there other reasons for making these traits `unsafe`?

romancardenas: The idea of making these traits `unsafe` is preventing users from implementing them over any structure. Ideally, only PACs via `svd2rust` should implement them. In this way, peripheral drivers etc. can rely over these traits to ensure that their APIs are safe even using raw pointers behind the scenes.

romancardenas (commented):
Changes done. Please, take a look and let me know what you think. I personally think that we can improve it considerably. I will comment on the PR with my alternative idea.

riscv/src/register/mcause.rs
```diff
@@ -30,10 +13,32 @@ pub enum Interrupt {
...
+    }
+
+    #[inline]
+    fn from_number(value: usize) -> Result<Self, usize> {
+        if value > 11 || value % 2 == 0 {
+            Err(value)
+        } else {
+            // SAFETY: valid interrupt number
+            unsafe { Ok(core::mem::transmute::<usize, Self>(value)) }
+        }
```
rmsyn, line 34: Could we change this to a:

```rust
match value {
    1 => Ok(Self::SupervisorSoft),
    // ...
    _ => Err(value),
}
```

Mainly to avoid the call to `transmute`, since we really don't need it.

riscv/src/register/mcause.rs
```diff
@@ -51,68 +56,52 @@ pub enum Exception {
...
-    fn try_from(value: Interrupt) -> Result<Self, Self::Error> {
-        match value {
-            Interrupt::Unknown => Err(Self::Error::Unknown),
-            _ => Ok(value as Self),
+    fn from_number(value: usize) -> Result<Self, usize> {
+        if value == 10 || value == 14 || value > 15 {
+            Err(value)
+        } else {
+            // SAFETY: valid exception number
+            unsafe { Ok(core::mem::transmute::<usize, Self>(value)) }
```
rmsyn, line 76: Same as above, try replacing with a `match` on expected values.

riscv/src/register/mcause.rs
romancardenas: The good thing about my `transmute` implementation is that on average it is faster (no need to have N cases, with just two comparisons you are ready to go)

riscv/src/register/mcause.rs
romancardenas: Same here, but with three comparisons

riscv/src/register/mcause.rs
rmsyn: The bad thing about `transmute` is that it is inherently unsafe (read "can introduce UB"), and should generally be avoided whenever possible.

Also, `transmute` is not necessarily faster:

```
from_match              time:   [147.28 ps 147.82 ps 148.47 ps]
                        change: [-2.0146% -0.1924% +1.7735%] (p = 0.86 > 0.05)
                        No change in performance detected.
Found 11 outliers among 100 measurements (11.00%)
  2 (2.00%) high mild
  9 (9.00%) high severe

     Running benches/transmute.rs (target/release/deps/transmute-b6412ad997b8e934)
Gnuplot not found, using plotters backend
from_transmute          time:   [150.79 ps 151.19 ps 151.64 ps]
                        change: [-0.1834% +2.5244% +4.7970%] (p = 0.04 < 0.05)
                        Change within noise threshold.
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild
  3 (3.00%) high severe
```

Running multiple iterations of `cargo bench` yields some cases where `transmute` is faster, but only slightly. The two implementations swap which is faster, sometimes `match` sometimes `transmute`. Always +/- ~4-5ps.

`match` has the advantage of always being safe.

```rust
#[repr(usize)]
#[derive(Copy, Clone, Debug, Eq, PartialEq)]
pub enum F {
    A = 1,
    B = 2,
    C = 3,
    D = 4,
}

impl F {
    pub fn from_num_transmute(val: usize) -> Option<Self> {
        if val == 0 || val > 4 {
            None
        } else {
            Some(unsafe { std::mem::transmute::<usize, Self>(val) })
        }
    }

    pub fn from_num_match(val: usize) -> Option<Self> {
        match val {
            1 => Some(Self::A),
            2 => Some(Self::B),
            3 => Some(Self::C),
            4 => Some(Self::D),
            _ => None,
        }
    }
}
```

The benchmark code:

```rust
// Using `match`
use criterion::{black_box, criterion_group, criterion_main, Criterion};
use ff::F;

fn bench_match(n: usize) {
    F::from_num_match(n);
}

fn criterion_benchmark(c: &mut Criterion) {
    c.bench_function("from_match", |b| b.iter(|| bench_match(black_box(20))));
}

criterion_group!(benches, criterion_benchmark);
criterion_main!(benches);
```

```rust
// Using `transmute
use criterion::{black_box, criterion_group, criterion_main, Criterion};
use ff::F;

fn bench_transmute(n: usize) {
    F::from_num_transmute(n);
}

fn criterion_benchmark(c: &mut Criterion) {
    c.bench_function("from_transmute", |b| b.iter(|| bench_transmute(black_box(20))));
}

criterion_group!(benches, criterion_benchmark);
criterion_main!(benches);
```
```rust

File: riscv/src/register/mcause.rs (hunk 5 above)
Comment (rmsyn) [reply riscv/src/register/mcause.rs:76]: > Same here, but with three comparisons

This is incorrect, the only valid exceptions are `0..=15`. This implementation allows any number that isn't `10`, `14`, or `15`.

What happens when someone calls:

```rust
Exception::from_number(0xbad)
```

rmsyn: > Same here, but with three comparisons

This is incorrect, the only valid `Exception` variants are `0..=15`. This implementation allows any number that isn't `10`, `14`, or `15`.

What happens when someone calls:

```rust
Exception::from_number(0xbad)
```

rmsyn: Same here, the output of three `cargo bench` runs:

```
    Running benches/match.rs (target/release/deps/match-570076475774c958)
Gnuplot not found, using plotters backend
from_match              time:   [114.02 ps 114.13 ps 114.23 ps]
                        change: [-25.459% -24.130% -23.185%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 11 outliers among 100 measurements (11.00%)
  4 (4.00%) low severe
  7 (7.00%) low mild

     Running benches/transmute.rs (target/release/deps/transmute-b6412ad997b8e934)
Gnuplot not found, using plotters backend
from_transmute          time:   [110.27 ps 110.48 ps 110.72 ps]
                        change: [-29.280% -27.967% -27.074%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) low severe
  1 (1.00%) low mild
```

---

```
     Running benches/match.rs (target/release/deps/match-570076475774c958)
Gnuplot not found, using plotters backend
from_match              time:   [113.96 ps 114.14 ps 114.30 ps]
                        change: [-0.6957% -0.2491% +0.2621%] (p = 0.32 > 0.05)
                        No change in performance detected.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) low severe
  1 (1.00%) low mild

     Running benches/transmute.rs (target/release/deps/transmute-b6412ad997b8e934)
Gnuplot not found, using plotters backend
from_transmute          time:   [110.88 ps 111.15 ps 111.37 ps]
                        change: [-0.3696% +0.0428% +0.5089%] (p = 0.85 > 0.05)
                        No change in performance detected.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) low mild
```

---

```
test result: ok. 0 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running benches/match.rs (target/release/deps/match-570076475774c958)
Gnuplot not found, using plotters backend
from_match              time:   [109.67 ps 109.85 ps 110.05 ps]
                        change: [-3.1358% -2.7367% -2.3282%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) low severe
  2 (2.00%) low mild

     Running benches/transmute.rs (target/release/deps/transmute-b6412ad997b8e934)
Gnuplot not found, using plotters backend
from_transmute          time:   [113.24 ps 113.52 ps 113.81 ps]
                        change: [+2.0011% +2.4700% +2.8800%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) low severe
  1 (1.00%) low mild
```

rmsyn: Same here, the output of three `cargo bench` runs:

```
     Running benches/match.rs (target/release/deps/match-570076475774c958)
Gnuplot not found, using plotters backend
from_match              time:   [112.89 ps 113.23 ps 113.57 ps]
                        change: [+1.7402% +2.1406% +2.5643%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) low severe
  1 (1.00%) low mild
  2 (2.00%) high mild

     Running benches/transmute.rs (target/release/deps/transmute-b6412ad997b8e934)
Gnuplot not found, using plotters backend
from_transmute          time:   [111.94 ps 112.12 ps 112.31 ps]
                        change: [-1.3954% -0.8420% -0.3073%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 11 outliers among 100 measurements (11.00%)
  3 (3.00%) low severe
  3 (3.00%) low mild
  1 (1.00%) high mild
  4 (4.00%) high severe
```

---

```
     Running benches/match.rs (target/release/deps/match-570076475774c958)
Gnuplot not found, using plotters backend
from_match              time:   [111.00 ps 111.12 ps 111.24 ps]
                        change: [-1.6529% -1.3156% -0.9716%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) low severe
  4 (4.00%) low mild
  1 (1.00%) high severe

     Running benches/transmute.rs (target/release/deps/transmute-b6412ad997b8e934)
Gnuplot not found, using plotters backend
from_transmute          time:   [111.62 ps 111.81 ps 112.00 ps]
                        change: [-0.6062% -0.0364% +0.5361%] (p = 0.90 > 0.05)
                        No change in performance detected.
Found 8 outliers among 100 measurements (8.00%)
  1 (1.00%) low severe
  3 (3.00%) low mild
  3 (3.00%) high mild
  1 (1.00%) high severe
```

---

```
Gnuplot not found, using plotters backend
from_match              time:   [113.22 ps 113.63 ps 113.99 ps]
                        change: [+0.1077% +0.8447% +1.5072%] (p = 0.02 < 0.05)
                        Change within noise threshold.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) low severe
  1 (1.00%) high mild

     Running benches/transmute.rs (target/release/deps/transmute-b6412ad997b8e934)
Gnuplot not found, using plotters backend
from_transmute          time:   [114.66 ps 115.01 ps 115.39 ps]
                        change: [+1.6948% +2.2350% +2.7761%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 5 outliers among 100 measurements (5.00%)
  2 (2.00%) low severe
  3 (3.00%) low mild
```

With 16 variants:

```rust
#[repr(usize)]
#[derive(Copy, Clone, Debug, Eq, PartialEq)]
pub enum F {
    A = 1,
    B = 2,
    C = 3,
    D = 4,
    E = 5,
    F = 6,
    G = 7,
    H = 8,
    I = 9,
    J = 10,
    K = 11,
    L = 12,
    M = 13,
    N = 14,
    O = 15,
    P = 16,
}

impl F {
    pub fn from_num_transmute(val: usize) -> Option<Self> {
        if val == 0 || val > 16 {
            None
        } else {
            Some(unsafe { std::mem::transmute::<usize, Self>(val) })
        }
    }

    pub fn from_num_match(val: usize) -> Option<Self> {
        match val {
            1 => Some(Self::A),
            2 => Some(Self::B),
            3 => Some(Self::C),
            4 => Some(Self::D),
            5 => Some(Self::E),
            6 => Some(Self::F),
            7 => Some(Self::G),
            8 => Some(Self::H),
            9 => Some(Self::I),
            10 => Some(Self::J),
            11 => Some(Self::K),
            12 => Some(Self::L),
            13 => Some(Self::M),
            14 => Some(Self::N),
            15 => Some(Self::O),
            16 => Some(Self::P),
            _ => None,
        }
    }
}
```

riscv/src/register/mcause.rs
rmsyn: This got me thinking about why the two implementations are so close in performance, and how the compiler translates to machine code.

Here is an example from [godbolt](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DEArgoKkl9ZATwDKjdAGFUtEywYgATAHZSjgBk8BkwAOXcAI0xiCQAOUgAHVAVCOwYXNw9vPySU2wEgkPCWKJiueKtMGzShAiZiAgz3T19LTGt8hlr6gkKwyOi4yzqGpqzW8x6%2B4tK4gEpLVBNiZHYOAFIvAGYwMHWAVgAhBlQAfXN0A4ARdY0AQVuH7YPD4kwE4ggzPAAvTDnro8EiYIgBqRjuUHYUHrHyHR6gxGgu4wrZXUFcUgIpGHVHorZY%2B5I0FOPGg/aElFI9HrNGgvzYxHQ2nogCclOJADEyVxMY9YTd7vznkcTqcWIZgPRAfdgWCqAxQVRiKgWKcCMRDAoWCYCJgIAx3CBQd8/nNQQBaWnQgDyCU6tKc2GtMLhjNBeCooMNLFR0N5rqcsNJPphXn2oK8eJZoI0rvhROJSNCAkw7oF4NoSnj7uJQlV%2BpMDAUTComHjoLQbxAIDYLBrGq1Or1NcdpswpCh1oN7nNAoBicRAv5PkFDyFIuOZwlRml%2BzHcqViuVqvFTAIyAQPfrJpSZstLrtDq2TpdsITVMREo3CG9kPPuaRXGjW2h%2BbYEGwNbuC0fiK2L5vgWn41ocv6DkmEYsi6776l%2BIBOOBl5Jj4gGgrBIEgFcSFJkirJoRh8HYDhuEYs%2B0GvuhwHwZyJG4acaEpiEHJIsOQqjvy9wAPRcaCACSXoAJ5LLu5brAAbBoErBBAAKSZ2WDIAYbwegQoJMAoMKSXKElxkQoJKOWhAeoqBAIOWSwEMCBAgI8PGgouCqgtJDCyRWAB0nmumOHALLQnD7LwnjcLwqCcAASmYakKEsKxidsPCkDZHBaHMCwANYgFsWzudleX5QV4n6JwkhBZooWcLwCggBoSXlQscCwEgaAsAkdDROQlAtW19AxMQXA%2BLEtU0LQerENVEAROVpARME9SCZwiWzcwxCCTaETaFUyWJS1bCCDaDC0AtKW8FgEQmMAThiFmi2nZgM5SmsWikPgbzVAAbpg1Unf4qhVLqayJcEer%2BT9tB4BEmqrS4WDTRqeD1iFpCfcQETJJgVz3ZK4NGPVfAGMACgAGp4JgADudqMLdMiCCIYjsFINPyEoajTbomIGLjpjmPoEPVZACyoPaaTfRapKkRLQsEBa9CfbQLIARaFojMAmBqcQeAKMg73iZIwDIBaRbpScZMMBaJwhBa7RUKFKMa1g/OyW0HRpA4DDOK4zR6IEwT9CUgyYrkqQCGMniB8kwcMNMAxlM7W01CMjSe1kmKVNUAjdA00f%2B7HkyjMnYfDFMvszAHCwxcsqwSH5AVlT9YUcKC3Mbhi7mDe5cafvgRDEGGWxcHMvDJalGVZTlBUT3lRWg6VpDBc9DdVTVdUnQ1MCICAlnWZ1EDde1xChKwazN8grft89mDdyQeDoHo/C06I4iM/fzMqOoP3s6QZOagkt01xwgU57TQbjaXU1lQSoC9CfM%2BsQO6gggC4Vq%2B8%2B4DyHvVUe2VcqTwnsVDgs954VQ4EvWqw9fL/y8HXBelUV4j2RtEFI9hJBAA%3D%3D):

```rust
#![no_std]

#[repr(usize)]
pub enum E {
    A = 1,
    B = 3,
    C = 5,
    D = 7,
    E = 9,
    F = 11,
}

#[no_mangle]
pub fn from_transmute(num: usize) -> Option<E> {
    if num > 11 || num % 2 == 0 {
        None
    } else {
        Some(unsafe { core::mem::transmute::<usize, E>(num) })
    }
}

#[no_mangle]
pub fn from_match(num: usize) -> Option<E> {
    match num {
        1 => Some(E::A),
        3 => Some(E::B),
        5 => Some(E::C),
        7 => Some(E::D),
        9 => Some(E::E),
        11 => Some(E::F),
        _ => None,
    }
}
```

---

```asm
; with rustc -C opt-level=3 --target riscv64gc-unknown-none-elf
from_transmute:
        sltiu   a1, a0, 12
        xori    a1, a1, 1
        andi    a2, a0, 1
        seqz    a2, a2
        or      a1, a1, a2
        addi    a1, a1, -1
        and     a0, a0, a1
        ret

from_match:
        addi    a0, a0, -1
        li      a1, 10
        bltu    a1, a0, .LBB1_2
        slli    a0, a0, 3
.Lpcrel_hi0:
        auipc   a1, %pcrel_hi(.Lswitch.table.from_match)
        addi    a1, a1, %pcrel_lo(.Lpcrel_hi0)
        add     a0, a0, a1
        ld      a0, 0(a0)
        ret
.LBB1_2:
        li      a0, 0
        ret

.Lswitch.table.from_match:
        .quad   1
        .quad   0
        .quad   3
        .quad   0
        .quad   5
        .quad   0
        .quad   7
        .quad   0
        .quad   9
        .quad   0
        .quad   11
```

riscv/src/register/mcause.rs
romancardenas: Soooo... looks like LLVM is smarter than I thought (again). I will move to the classic, safe match fashion ASAP :) Thank you so much!

romancardenas:
Thanks for the feedback @rmsyn !

I implemented my alternative approach in the [`riscv-pac2` branch](https://github.com/rust-embedded/riscv/tree/riscv-pac2). Please take a look and let me know what you think. Personally, I feel like it is more user friendly, and somehow it makes sense to me that the `Interrupt` and `Exception` enums are in `riscv::interrupt` instead of `riscv::register`

rmsyn:
> it makes sense to me that the Interrupt and Exception enums are in riscv::interrupt instead of riscv::register

That makes sense to me, too.

Really either one can work, because the values are read from the `mcause` and `scause` registers (which makes sense for `riscv::register`).

`riscv::interrupt` makes obvious sense because `Interrupt` and `Exception` are explicitly about interrupts.

:shrug:

rmsyn (approved):
I probably need more time to review the proc-macro stuff, but it looks good to me.

rmsyn:
> I implemented my alternative approach in the [riscv-pac2 branch](https://github.com/rust-embedded/riscv/tree/riscv-pac2).

After an initial review, I do like the code organization in `riscv-pac2` more. Mostly aesthetic preferences, but it does feel like it clarifies things a bit.

The unification of one `Trap` type is also nice from an ergonomics perspective regarding the `mcause`/`scause` registers.

romancardenas:
So I think the new `riscv` version looks good (still needs some testing etc.) BUT... it's time for `riscv-rt` and the interrupt/exception handlers. We need to provide an out-of-the-box solution for dealing with custom interrupts and exceptions. I guess that it's time for thinking on macros that:

- declare the interrupt and exception handlers as `extern "C" fn`
- define `__INTERRUPTS` and `__EXCEPTIONS` arrays from the enums
- In vectored mode, creates the vector table etc.

romancardenas:
I found some time to keep working on this. In the new commit, the following code:

```rust
#[pac_enum(unsafe PriorityNumber)]
#[derive(Clone, Copy, Debug, Eq, PartialEq)]
enum Priority {
    P0 = 0,
    P1 = 1,
    P2 = 2,
    P3 = 3,
}
```

expands to:

```rust
#[derive(Clone, Copy, Debug, Eq, PartialEq)]
enum Priority {
    P0 = 0,
    P1 = 1,
    P2 = 2,
    P3 = 3,
}
unsafe impl riscv_pac::PriorityNumber for Priority {
    const MAX_PRIORITY_NUMBER: u8 = 3;
    #[inline]
    fn number(self) -> u8 {
        self as _
    }
    #[inline]
    fn from_number(number: u8) -> Result<Self, u8> {
        match number {
            3 => Ok(Self::P3),
            0 => Ok(Self::P0),
            1 => Ok(Self::P1),
            2 => Ok(Self::P2),
            _ => Err(number),
        }
    }
}
```

On the other hand, the following code:

```rust
#[pac_enum(unsafe CoreInterruptNumber)]
#[derive(Clone, Copy, Debug, Eq, PartialEq)]
enum Interrupt {
    I1 = 1,
    I2 = 2,
    I4 = 4,
    I7 = 7,
}
```

Expands to:

```rust
#[derive(Clone, Copy, Debug, Eq, PartialEq)]
enum Interrupt {
    I1 = 1,
    I2 = 2,
    I4 = 4,
    I7 = 7,
}
unsafe impl riscv_pac::InterruptNumber for Interrupt {
    const MAX_INTERRUPT_NUMBER: usize = 7;
    #[inline]
    fn number(self) -> usize {
        self as _
    }
    #[inline]
    fn from_number(number: usize) -> Result<Self, usize> {
        match number {
            2 => Ok(Self::I2),
            1 => Ok(Self::I1),
            4 => Ok(Self::I4),
            7 => Ok(Self::I7),
            _ => Err(number),
        }
    }
}
unsafe impl riscv_pac::CoreInterruptNumber for Interrupt {} // it also implements the marker trait

// It declares the existence of all the interrupt handlers
extern "C" {
    fn I2();
    fn I1();
    fn I4();
    fn I7();
}

// It generates the array with handlers
#[no_mangle]
pub static __CORE_INTERRUPTS: [Option<unsafe extern "C" fn()>; 7 + 1] = [
    None,
    Some(I1),
    Some(I2),
    None,
    Some(I4),
    None,
    None,
    Some(I7),
];

// It implements the interrupt dispatching function
#[no_mangle]
unsafe extern "C" fn _dispatch_core_interrupt(code: usize) {
    extern "C" {
        fn DefaultHandler();

    }
    if code < __CORE_INTERRUPTS.len() {
        let h = &__CORE_INTERRUPTS[code];
        if let Some(handler) = h {
            handler();
        } else {
            DefaultHandler();
        }
    } else {
        DefaultHandler();
    }
}
```

It should work as is for direct mode. I still need to adapt this solution to vectored mode.

romancardenas:
It already creates the interrupt vector table. However, I think we should discuss the new approach with the @rust-embedded/tools team to align `svd2rust` with the new structure

riscv-rt/src/lib.rs
```diff
@@ -520,26 +522,24 @@ pub struct TrapFrame {
...
+            let trap_frame = &*trap_frame;
+            if code < __EXCEPTIONS.len() {
+                let h = &__EXCEPTIONS[code];
+                if let Some(handler) = h {
+                    handler(trap_frame);
+                } else {
+                    ExceptionHandler(trap_frame);
+                }
             } else {
                 ExceptionHandler(trap_frame);
```
rmsyn, line 540:
```suggestion
        xcause::Trap::Exception(code) if code < __EXCEPTIONS.len() => {
            match __EXCEPTIONS[code].as_ref() {
                Some(handler) => handler(&*trap_frame),
                None => ExceptionHandler(&*trap_frame),
            }
        }
        xcause::Trap::Exception(_) => ExceptionHandler(&*trap_frame),
```

riscv/Cargo.toml
```diff
@@ -26,3 +26,4 @@ critical-section-single-hart = ["critical-section/restore-state-bool"]
 [dependencies]
 critical-section = "1.1.2"
 embedded-hal = "1.0.0"
+riscv-pac = { path = "../riscv-pac", version = "0.1.2", default-features = false }
```
rmsyn, line 29: Do we want to add an extra `feature` to the `riscv` crate to enable the vectors feature in `riscv-pac`?

rmsyn (approved):
Minor style nit, and a quick question about a exposing `riscv-pac` features in `riscv`.

LGTM

riscv/Cargo.toml
romancardenas: So far, `riscv-pac` only has a default feature to opt out of procedural macros (i.e., `syn`, `proc-macro2`, and `quote` dependencies). However, I don't know if this effort is worthy it, as in the end, you will end up using `riscv-rt`, which depends on procedural macros.

romancardenas: maybe we could do something like:
```
[features]
default = ["riscv-pac/default"]
```

riscv-pac/macros/src/lib.rs
```diff
@@ -0,0 +1,367 @@
...
+                vectors.push(quote! { None });
+            }
+        }
+        vectors
+    }
+
+    fn vector_table(&self) -> TokenStream2 {
+        let mut asm = String::from(
+            r#"
+core::arch::global_asm!("
```
rmsyn, line 198:
```suggestion
#[cfg(any(target_arch = "riscv32", target_arch = "riscv64"))]
core::arch::global_asm!("
```

Fixes the `ui` tests.

romancardenas:
Exciting times! I think the `riscv` ecosystem will improve a lot with #222 and this PR.

I will wait until #222 is solved and merge the outcome, so we can adapt the work to the new approach.
In the meantime, I will keep updating my forks for `svd2rust` so we have an out-of-the-box, easy-to-apply tool to help RISC-V PACs adopt our new ecosystem.

romancardenas:
I updated this PR to use the new features coming from #222 .
An additional change is that I moved the macros to `riscv` instead of `riscv-pac`. As now `riscv-pac` is fully re-exported into `riscv`, it makes more sense, specially when leading with `svd2rust` (I'm also working on that).

This branch is quite a mess. I will divide it into small PRs so it will be easier to review.

PS: Rust nightly is complaining about a bug in LLVM for inline assembly with x86 targets. I think it is a matter of time that the error message dissapears: https://github.com/rust-lang/rust/pull/127935

rmsyn:
> Rust nightly is complaining about a bug in LLVM for inline assembly with x86 targets

One potential way to solve that issue before the fix makes it into a nightly release could be to feature-gate all assembly with:

 ```rust
#[cfg(any(target_arch = "riscv32", target_arch = "riscv64"))]
...
```

romancardenas:
The thing is that it still complains for RISC-V targets :/

rmsyn:
> The thing is that it still complains for RISC-V targets :/

What about switching the label to `1b:`, `b1:`, or something similar? Not sure about any concern for label collision.

romancardenas:
fixed

rmsyn (approved):

romancardenas:
Closing this PR in favor of #223 , which has a cleaner commit history and cooler macros for defining interrupt and exception handlers

Issue #212, opened 2024-05-21
`riscv`: Consider strategy for exception safe code


rmsyn:
From the [Rustinomicon](https://doc.rust-lang.org/nomicon/exception-safety.html) section on exception safety:

```
In unsafe code, we must be exception safe to the point of not violating memory safety. 

We'll call this minimal exception safety.
```

There are a number of locations in the code that use explicit panics (e.g. `unimplemented!`, `assert.*!` macros). Sometime, these `panic`ing calls are made within `safe` functions, which users may call inside an `unsafe` context.

`riscv` should convert all explicit and implicit `panic`ing code into fallible code that returns an `Option` or `Result`.

For example, `Mcounteren::hpm` could be changed from:

```rust
    /// Supervisor "hpm\[x\]" Enable (bits 3-31)
    #[inline]
    pub fn hpm(&self, index: usize) -> bool {
        assert!((3..32).contains(&index));
        self.bits & (1 << index) != 0
    }
```

To the fallible:

```rust
    /// Supervisor "hpm\[x\]" Enable (bits 3-31)
    #[inline]
    pub fn hpm(&self, index: usize) -> Option<bool> {
        if (3..32).contains(&index) {
            Some(self.bits & (1 << index) != 0)
        } else {
            None
        }
    }
```

Any strategy will almost definitely require breaking changes, and a version bump.

rmsyn:
> I would like to discuss this with @rust-embedded/cortex-m and get to a similar design to make the developing experience more homogeneous.

Should we add this as a talking point for the next `rust-embedded` WG meeting?

romancardenas:
Sure! I didn't add it to this week's meeting because I'm currently in a conference, but we can give it a try next week.

rmsyn:
In the last WG meeting, we discussed some trade-offs of having explicit `panic`s in the code.

Some advantages of explicit panics are:

- indicating conditions where we convey "this should never happen".
- panics also ensure a "fail-fast" condition for exceptional code where the user can't really do anything to handle the error.
- panic handlers can also use the passed in context to print out info, call subsequent specialized handlers, etc.

Some advantages of fallible functions:

- can take advantage of Rust facilities to bubble-up errors
- error handling looks more like the rest of Rust
- defers the decision to panic to the user (less opinionated)

Also, since we are discussing breaking changes, we should also consider using Rust's version of `static_assert`, e.g.:

```rust
impl Mcounteren {
// ...
[...]

PR #213, opened 2024-05-21
riscv: add `mcountinhibit` module

rmsyn:
Adds the `riscv::register::mcountinhibit` module for the `mcountinhibit` CSR.

### DIFF
--- a/riscv/src/register/mcountinhibit.rs
+++ b/riscv/src/register/mcountinhibit.rs
@@ -1,0 +1,116 @@
+//! `mcountinhibit` register
+
+use crate::bits::{bf_extract, bf_insert};
+
+/// `mcountinhibit` register
+#[derive(Clone, Copy, Debug)]
+pub struct Mcountinhibit {
+    bits: usize,
+}
+
+impl Mcountinhibit {
+    /// Machine "cycle\[h\]" Disable
+    #[inline]
+    pub fn cy(&self) -> bool {
+        bf_extract(self.bits, 0, 1) != 0
+    }
+
+    /// Sets whether to inhibit the "cycle\[h\]" counter.
+    ///
+    /// Only updates the in-memory value, does not modify the `mcountinhibit` register.
+    #[inline]
+    pub fn set_cy(&mut self, cy: bool) {
+        self.bits = bf_insert(self.bits, 0, 1, cy as usize);
+    }
+
+    /// Machine "instret\[h\]" Disable
+    #[inline]
+    pub fn ir(&self) -> bool {
+        bf_extract(self.bits, 2, 1) != 0
+    }
+
+    /// Sets whether to inhibit the "instret\[h\]" counter.
+    ///
+    /// Only updates the in-memory value, does not modify the `mcountinhibit` register.
+    #[inline]
+    pub fn set_ir(&mut self, ir: bool) {
+        self.bits = bf_insert(self.bits, 2, 1, ir as usize);
+    }
+
+    /// Machine "hpm\[x\]" Disable (bits 3-31)
+    #[inline]
+    pub fn hpm(&self, index: usize) -> bool {
+        assert!((3..32).contains(&index));
+        bf_extract(self.bits, index, 1) != 0
+    }
+
+    /// Sets whether to inhibit the "hpm\[X\]" counter.
+    ///
+    /// Only updates the in-memory value, does not modify the `mcountinhibit` register.
+    #[inline]
+    pub fn set_hpm(&mut self, index: usize, hpm: bool) {
+        assert!((3..32).contains(&index));
+        self.bits = bf_insert(self.bits, index, 1, hpm as usize);
+    }
+}
+
+read_csr_as!(Mcountinhibit, 0x320);
+write_csr_as!(Mcountinhibit, 0x320);
+set!(0x320);
+clear!(0x320);
+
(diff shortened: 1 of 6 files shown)


riscv/src/register/mcountinhibit.rs
```diff
@@ -0,0 +1,95 @@
...
+    /// Only updates the in-memory value, does not modify the `mcountinhibit` register.
+    #[inline]
+    pub fn set_ir(&mut self, ir: bool) {
+        self.bits = bf_insert(self.bits, 2, 1, ir as usize);
+    }
+
+    /// Machine "hpm\[x\]" Disable (bits 3-31)
+    #[inline]
+    pub fn hpm(&self, index: usize) -> Option<bool> {
+        if (3..32).contains(&index) {
```
romancardenas, line 43: Shouldn't this conditional be the other way around?

riscv/src/register/mcountinhibit.rs
```diff
@@ -0,0 +1,95 @@
...
+            Some(bf_extract(self.bits, index, 1) != 0)
+        }
+    }
+
+    /// Sets whether to inhibit the "hpm\[X\]" counter.
+    ///
+    /// Only updates the in-memory value, does not modify the `mcountinhibit` register.
+    #[inline]
+    pub fn set_hpm(&mut self, index: usize, hpm: bool) -> Option<()> {
+        if (3..32).contains(&index) {
```
romancardenas, line 55: same

riscv/src/register/mcountinhibit.rs
```diff
@@ -0,0 +1,95 @@
...
+        } else {
+            Some(bf_extract(self.bits, index, 1) != 0)
+        }
+    }
+
+    /// Sets whether to inhibit the "hpm\[X\]" counter.
+    ///
+    /// Only updates the in-memory value, does not modify the `mcountinhibit` register.
+    #[inline]
+    pub fn set_hpm(&mut self, index: usize, hpm: bool) -> Option<()> {
```
romancardenas, line 54: I feel like a `Result<(), ()>` is more suitable than an `Option<()>`.

riscv/src/register/mcountinhibit.rs
```diff
@@ -0,0 +1,95 @@
...
+/// Machine cycle Disable
+    , set_cy, clear_cy, 1 << 0);
+
+set_clear_csr!(
+/// Machine instret Disable
+    , set_ir, clear_ir, 1 << 2);
+
+#[inline]
+pub unsafe fn set_hpm(index: usize) -> Option<()> {
+    if (3..32).contains(&index) {
```
romancardenas, line 79: same

riscv/src/register/mcountinhibit.rs
```diff
@@ -0,0 +1,95 @@
...
+        None
+    } else {
+        _set(1 << index);
+        Some(())
+    }
+}
+
+#[inline]
+pub unsafe fn clear_hpm(index: usize) -> Option<()> {
+    if (3..32).contains(&index) {
```
romancardenas, line 89: same

romancardenas (changes requested):
Thanks for the PR!

About returning a result/option: I think it is better if we `assert` for now and discuss in #212 about transitioning to a panic-less strategy

riscv/src/register/mcountinhibit.rs
```diff
@@ -0,0 +1,95 @@
+//! `mcountinhibit` register
+
+use crate::bits::{bf_extract, bf_insert};
+
+/// `mcountinhibit` register
+#[derive(Clone, Copy, Debug)]
+pub struct Mcountinhibit {
+    bits: usize,
```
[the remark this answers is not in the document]
rmsyn: I tried to make `bits` a `u32`, but the `read!`/`write!` macros need a `usize`.

riscv/src/register/mcountinhibit.rs
rmsyn: Definitely, especially since the compiler will warn about unused `Result`. I wasn't sure of the error type for the crate, so reverted to `Option`.

I can change it to `Result` in the fixup set of changes.

rmsyn: `clippy` fails with the `Result<_, ()>` return type:

```
error: this returns a `Result<_, ()>`
  --> riscv/src/register/mcountinhibit.rs:42:5
   |
42 |     pub fn hpm(&self, index: usize) -> Result<bool, ()> {
```

I know you mentioned keeping the `assert` until we address the general approach to fallible function, so I'll probably just revert to using those for now.

If we do decide to use `Result` types, we should probably have a custom `Error` type for the crate, as well.

romancardenas (dismissed):
LGTM

romancardenas:
Clippy is complaining in merge queue: https://github.com/rust-embedded/riscv/actions/runs/9201547297

@rmsyn  please can you update your PR?

rmsyn:
> Clippy is complaining in merge queue: https://github.com/rust-embedded/riscv/actions/runs/9201547297
> 
> @rmsyn please can you update your PR?

Did something change from the PR CI runs? I fixed the comment clippy was unhappy with, but it was unrelated to my code changes...

romancardenas (approved):
LGTM

seems that nightly has new lints that previously didn't trigger

PR #217, opened 2024-05-25
riscv: add `mcounteren` in-memory update functions

rmsyn:
Adds functions for updating the in-memory `Mcounteren` structure without touching the CSR.

Allows manipulating all bitfields in-memory, and writing the entire value back to the CSR at once.


romancardenas (approved):
LGTM!

It would be interesting if we are able to update all the registers using brand new macros that follow this fashion.

rmsyn:
> It would be interesting if we are able to update all the registers using brand new macros that follow this fashion.

I can work on something. Would probably want at least two versions of the macro, `read_only_csr` and `read_write_csr` (or whatever naming scheme).

Might also be good to export the macros for downstream users with custom CSRs on their SoC.

Issue #218, opened 2024-05-25
`riscv`: Add macro to define CSR register types


rmsyn:
As discussed in https://github.com/rust-embedded/riscv/pull/217, it would be nice to have a macro to define CSR register types.

I worked on a couple solutions:

- macros using the `paste` crate: <https://github.com/rust-embedded/riscv/compare/master...rmsyn:riscv:riscv/csr-macro-paste>
  - less verbose
  - more consistent docs across definitions
  - requires adding the `paste` crate as a dependency, and exposing the `paste!` macro as a public re-export 
- macros without using the `paste` crate: <https://github.com/rust-embedded/riscv/compare/master...rmsyn:riscv:riscv/csr-macro-no-paste>
  - more verbose
    - requires caller-supplied docstrings and `set_<field>` values
    - can potentially omit the docstrings at the expense of having generic documentation, e.g. `/// Gets the field value.`
  - potentially less consistent docs across definitions
  - requires no additional dependencies

What are some thoughts and/or preferences about the two implementations?

rmsyn:
We could also potentially add a `mask` parameter to describe the bitmask of readable/writable fields (or two masks if those sets are not identical).

Adding a `mask` parameter(s) would allow us to add `CsrType::bits` and `CsrType::set` methods to read and write legal values safely. We could also then safely provide a `impl From<usize> for CsrType` implementation that calls `CsrType::set`.

romancardenas:
I was thinking more about `csr_field_read`, `csr_field_write`, and `csr_field_read_write` macros that would be called independently for every field. Maybe this is not *that* worthy for having macros, though...

Regarding option 1: I would prefer to leave as less dependencies as possible in `riscv`

Regarding the `mask` parameter: I am not sure what you mean. Is it a parameter of the macro? We also need to think about WARL registers, in which you may read something different from what you wrote.

rmsyn:
>  I would prefer to leave as less dependencies as possible in riscv

Agreed, I just included it because it reduces the verbosity of the macro call.

> Regarding the mask parameter: I am not sure what you mean. Is it a parameter of the macro?

Yes, it would be a macro parameter to define the valid bitmask for the register.

For example, a register with valid fields bitranges `[7:7]`, `[1:1]`, `[0:0]`, the `mask` parameter would be `0b1000_0011` or `0x83`.

Which would then allow us to provide a safe interfaces for creating/setting all fileds at once:

```rust
[...]

PR #219, opened 2024-05-30
riscv: add CSR-defining macros

rmsyn:
Adds helper macros for defining CSR types:

- `read-write`:
  - have both getter and setter methods
  - able to read and write the CSR
- `read-only`:
  - only has getter methods
  - able to read the CSR
- `write-only`:
  - only has setter methods
  - able to write the CSR

Related to:  https://github.com/rust-embedded/riscv/issues/218


riscv/src/register/tests.rs
```diff
@@ -0,0 +1,25 @@
...
+    set_multi_range: 1..=3,
+}
+
+read_write_csr_field!(
+    Mtest,
+    "multi-bit field",
+    multi_field: [4:7],
+    "setter multi-bit field",
+    set_multi_field: [4:7],
+);
```
rmsyn, line 25: Test calls to the new macros. The above is just a basic compile test.

Still need to add unit tests, and macro calls for the `read-only` and `write-only` variants.

riscv/src/register/macros.rs
```diff
@@ -397,3 +397,446 @@ macro_rules! clear_pmp {
...
+                Self { bits: bits & $mask }
+            }
+
+            /// Gets the raw bits value.
+            pub const fn bits(&self) -> usize {
+                self.bits & $mask
+            }
+
+            /// Gets the bitmask for the CSR type's bitfields.
+            pub const fn bitmask(&self) -> usize {
```
romancardenas, line 684: Any reason not to prune this and just use `Self::BITMASK`?

romancardenas (commented):
Kind of like it! the mask idea is very nice. I miss two main things:

- Less repetition in macros. Maybe we could atomize them and leave one "core" macro that creates the struct with bitmask and from_bits/bits, another for the get/set... 
- Using custom types (mainly enums) for setters and getters. That would be awesome (for example [XLEN](https://github.com/rust-embedded/riscv/blob/7291ac280ce9af03878c0ad7eaa7c31f2acce90f/riscv/src/register/misa.rs#L13)

riscv/src/register/macros.rs
rmsyn: The main reason I included it is when you have:

```rust
let csr = SomeCsrType::from_bits(0);
let what_is_the_bitmask_again = csr.bitmask();
```

Or in situations where referring to the specific type isn't important. I guess this wouldn't be all that relevant unless we implemented a trait for CSR types.

For example, a collection of CSR types where the concrete types differ, but the user still wants to access the bitmask of each item. In that case `CsrType::BITMASK` is less ergonomic than `csr_var.bitmask`.

rmsyn: The main reason I included it is when you have:

```rust
let csr = SomeCsrType::from_bits(0);
let what_is_the_bitmask_again = csr.bitmask();
```

Or in situations where referring to the specific type isn't important. I guess this wouldn't be all that relevant unless we implemented a trait for CSR types.

For example, a collection of CSR types where the concrete types differ, but the user still wants to access the bitmask of each item. In that case, `CsrType::BITMASK` is less ergonomic than `csr_var.bitmask()`.

riscv/src/register/macros.rs
```diff
@@ -397,3 +397,446 @@ macro_rules! clear_pmp {
...
+        $(
+            $crate::read_write_csr_field!(
+                $ty,
+                $field_doc,
+                $field: $get_bit,
+                $set_field_doc,
+                $set_field: $set_bit,
+            );
+        )+
+    };
```
rmsyn, line 428: > Less repetition in macros. Maybe we could atomize them and leave one "core" macro that creates the struct with bitmask and from_bits/bits, another for the get/set...

That's essentially what is there, each of the branches for the various ways to call `[read,write]_[only,write]_csr` call the internal branch for creating the struct.

Then, the repeated part (in the `$(...)+`) creates the getters/setters. I guess we could also make `read_write_csr_field` into another branch of the macro...

Other than that, I'm not sure exactly what you're asking for. What changes would you like to see?

rmsyn:
> Using custom types (mainly enums) for setters and getters.

We could add another branch that takes an `$var_ty: ident, [$($variant:ident, $variant_val:literal),*]$(,)`. That would allow to define a custom `enum` type, and use that `enum` in the getters/setters.

I can try to work something out.

riscv/src/register/macros.rs
romancardenas: What I mean is that, for example, [this](https://github.com/rust-embedded/riscv/blob/fd163b109404372983ef0fa82fec518d9ccac2d4/riscv/src/register/macros.rs#L571) part of the macro where the register struct is defined appears three times. It would be nice to have it somehow just once in a separate, basic macro and, depending on the macro branches, call this macro

rmsyn: Ah, I see what you mean. Because each of the macros has this helper branch, make it into its own macro that each one could call.

Alright, I'll work on separating it out. Maybe with a note that you can, but probably don't want to, call this helper macro on its own.

rmsyn:
> I can try to work something out.

I added the `enum` branches, and an extra helper macro.

Let me know what you think.

romancardenas:
@rmsyn I'm on it, but macros are difficult to review... 😥 

Looks great. The only thing now is that I prefer to leave some macros out of the equation, and then let us call the proper sequence of macros according to the CSR. I think this will help us in the future to deal with macros (modular vs monolithic).

Regarding `from` and `into`: if we can get help from `From` and `TryFrom` traits, I'd go for them instead of maintaining repetitive methods.

Great job!! I'll keep studying all your code 😁

riscv/src/register/macros.rs
```diff
@@ -397,3 +397,681 @@ macro_rules! clear_pmp {
...
+                 match val {
+                     $($value => Some(Self::$variant),)+
+                     _ => None,
+                 }
+             }
+
+             /// Converts the variant into a [`usize`].
+             pub const fn into_usize(self) -> usize {
+                 self as usize
+             }
```
rmsyn, line 474: > Regarding from and into: if we can get help from From and TryFrom traits, I'd go for them instead of maintaining repetitive methods.

While I agree cutting down on repetition is good, adding these `from`/`into` methods here actually provides utility. Because [const traits](https://github.com/rust-lang/rust/issues/67792) aren't available yet, these methods allow for constructing CSR registers in `const` contexts.

We could probably workshop the name (e.g. `from_inner`), but I do think these functions are worth keeping.

Changing the inner representation type to a macro-defined type might also be a good thing to add.

If it's a deal-breaker for merging the PR, I can remove them.

riscv/src/register/macros.rs
```diff
@@ -397,3 +397,681 @@ macro_rules! clear_pmp {
...
+#[macro_export]
+macro_rules! read_write_csr {
+    ($doc:expr,
+     $ty:ident: $csr:tt,
+     $mask:tt,
+     $(
+     $field_doc:expr,
+     $field:ident: $get_bit:literal,
+     $set_field_doc:expr,
+     $set_field:ident: $set_bit:literal$(,)?
```
rmsyn, line 505: Another thing I would like your opinion on @romancardenas, do you think there are use-cases where users would want to define different bits for the `get` and `set` functions?

I couldn't really think of a place where this would be a good idea, so maybe the `bit` argument could be a single parameter?

Same idea for the range and multi-bit variants.

rmsyn:
Updated the changes onto the latest HEAD commit.

Squashed the previous fixup commit to simplify review.

Added the necessary changes to incorporate the addition of fallible functions, and changed the doc-string arguments.

Let me know if you like the changes.

I still think the bit-field and bit-range macro arguments could be simpler, especially now with four versions of essentially the same macro argument (e.g. `$field, $try_field, $set_field, $try_set_field`). 

Currently, it feels like having the user supply the same argument past the colon all four times is a potential footgun.

Instead, we could change the `$field:  $field_bit` to `$field` for all four variants, and add a single `$bit_arg: $bit` at tthe end. Thoughts?

rmsyn:
@romancardenas I added some more fixes + improvements to the macro helpers. Reduced some redundancy, and hopefully made the macro invocations more readable.

Please let me know what you think.

More or less, this PR is probably pretty close to "finished"(tm).

Let me know if there are any more changes you would like to see.

romancardenas:
I'll try to review this later today. Thanks!

romancardenas:
Looks good, but it is true that having to do something like:

```rust
    field: {
        /// Gets the `cycle[h]` inhibit field value.
        cy,
        /// Attempts to get the `cycle[h]` inhibit field value.
        try_cy,
        /// Sets the `cycle[h]` inhibit field value.
        ///
        /// **NOTE**: only updates the in-memory value without touching the CSR.
        set_cy,
        /// Attempts to set the `cycle[h]` inhibit field value.
        ///
        /// **NOTE**: only updates the in-memory value without touching the CSR.
        try_set_cy,
        bit: 0,
    },
```

is perhaps too verbose. Can't we do something like:

```rust
field: {
        /// `cycle[h]` inhibit field value
        cy,
        bit: 0,
    },
```

And then generate the same code? Just following the `try` and `set` notation.

Also, I feel like there are a lot of branches with similar code. Can't we simplify the macros using recursive calls that match patterns, consuming part of the tokens? I think in this way we would have simpler pattern matching and less duplicates.

rmsyn:
> Looks good, but it is true that having to do something like:
> 
> ```rust
>     field: {
>         /// Gets the `cycle[h]` inhibit field value.
>         cy,
>         /// Attempts to get the `cycle[h]` inhibit field value.
>         try_cy,
>         /// Sets the `cycle[h]` inhibit field value.
>         ///
>         /// **NOTE**: only updates the in-memory value without touching the CSR.
>         set_cy,
>         /// Attempts to set the `cycle[h]` inhibit field value.
>         ///
>         /// **NOTE**: only updates the in-memory value without touching the CSR.
>         try_set_cy,
>         bit: 0,
>     },
> ```
> 
> is perhaps too verbose. Can't we do something like:
> 
> ```rust
> field: {
>         /// `cycle[h]` inhibit field value
>         cy,
>         bit: 0,
>     },
> ```
>
> And then generate the same code? Just following the `try` and `set` notation.

Yes, this is what introducing the `paste` dependency solves. It requires `proc_macros` at the very least, and rewriting at least part of the `paste` crate locally, if we don't import it.

Personally, using the crate dependency is the better option, as it is actively maintained + high code quality.

I really do think the use-case justifies the additional dependency.

I'll add a commit using the `paste` crate.

> Also, I feel like there are a lot of branches with similar code. Can't we simplify the macros using recursive calls that match patterns, consuming part of the tokens? I think in this way we would have simpler pattern matching and less duplicates.

Could you make comments on the specific places where you think they could be refactored into recursive branches?

I have a couple of those branches in the `read_only_csr_field` and `write_only_csr_field` to allow for re-use in `read_write_csr`. Where else can be simplified in a similar way?

riscv/src/register/macros.rs
```diff
@@ -600,3 +600,979 @@ macro_rules! clear_pmp {
...
+                            index: $field_start,
+                            min: $field_start,
+                            max: $field_end,
+                        })
+                     }
+                 }
+             }
+         }
+    };
+}
```
rmsyn, line 1400: I rewrote the `read_only_csr_field` using the `paste` macro.

Please let me know if you are alright adding the dependency, and I'll rewrite the rest of the macros.

Please be sure about this decision, as it is a **lot** of work rewriting these macros.

riscv/src/register/tests/read_only_csr.rs
```diff
@@ -0,0 +1,133 @@
...
+        MtestFieldEnum {
+            range: [7:11],
+            default: Field1,
+            Field1 = 1,
+            Field2 = 2,
+            Field3 = 3,
+            Field4 = 15,
+        }
+    }
+);
```
rmsyn, line 37: You can see how much the `paste` crate simplifies macro invocation, it is substantial.

The savings are even greater for the `read_write_csr*` variants.

riscv/src/register/macros.rs
romancardenas: Then let's cool this down until we resolve other design decisions. I can say that, as of now, I like how it looks

romancardenas:
About reducing code size. Let's take only two pattern branches in `read_only_csr_field`:

```rust
#[macro_export]
macro_rules! read_only_csr {
    ($(#[$doc:meta])+
     $ty:ident: $csr:tt,
     mask: $mask:tt,
     $(
     $(#[$field_doc:meta])+
     $field:ident: $bit:literal$(,)?
     )+
    ) => {
        $crate::csr! { $(#[$doc])+ $ty: $csr, $mask }

        $(
            $crate::read_only_csr_field! {
                $ty,
                $(#[$field_doc])+
                $field: $bit,
            }
        )+

        $crate::read_csr_as!($ty, $csr);
    };

    ($(#[doc:meta])+
     $ty:ident: $csr:tt,
     mask: $mask:tt,
     $(
     $(#[$field_doc:meta])+
     $field:ident: $bit:literal ..= $bit_end:literal$(,)?
     )+) => {
        $crate::csr! { $(#[$doc])+ $ty: $csr, $mask }

        $(
            $crate::read_only_csr_field! {
                $ty,
                $(#[$field_doc])+
                $field: $bit ..= $bit_end,
            }
        )+

        $crate::read_csr_as!($ty, $csr);
    };
    ...
}
```

Both branches do something like:

```rust
        $crate::csr! { $(#[$doc])+ $ty: $csr, $mask }

        $(
            $crate::read_only_csr_field! {
                $ty,
                $(#[$field_doc])+
                <THIS LINE IS THE ONLY THING THAT CHANGES>
            }
        )+

        $crate::read_csr_as!($ty, $csr);
```

Looking at this specific example, the macro calls other macros (`csr`, `read_csr_as`, and `read_only_csr_field` a few times). So what is the benefit of having a single macro with many pattern branches that just propagate to other macro calls? Why don't we *keep it simple* and call in the user code to the `read_only_csr_field` all the times we want? In this way, we keep all macros simpler and users call them as many times as required.

riscv/src/register/macros.rs
```diff
@@ -600,3 +600,979 @@ macro_rules! clear_pmp {
...
+                if $bit < max_width {
+                    self.bits = $crate::bits::bf_insert(self.bits, $bit, 1, $field as usize);
+                    Ok(())
+                } else {
+                    Err($crate::result::Error::IndexOutOfBounds {
+                        index: $bit,
+                        min: 0,
+                        max: max_width,
+                    })
+                }
```
romancardenas, line 1432: We can do this as a compile-time check, right? we do know the value of `max_width` and `$bit`, so we can do a compile-time check to verify that everything is OK and just use the `pub fn $field` function, as we are sure that it will not fail.

riscv/src/register/macros.rs
```diff
@@ -600,3 +600,979 @@ macro_rules! clear_pmp {
...
+            $(#[$try_field_doc])+
+            #[inline]
+            pub fn $try_field(&mut self, index: usize, $field: bool) -> $crate::result::Result<()> {
+                let max_width = core::mem::size_of_val(&self.bits) * 8;
+
+                if index < max_width && ($bit_start..=$bit_end).contains(&index) {
+                    self.bits = $crate::bits::bf_insert(self.bits, index, 1, $field as usize);
+                    Ok(())
+                } else {
+                    Err($crate::result::Error::IndexOutOfBounds {
```
romancardenas, line 1459: Here more or less the same. We need this `try_field` method, as we don't know if the index will be valid. But we can be sure at compile time that `$bit_start` and `$bit_end` are correct, so we just need to check that `index` is inside this range.

romancardenas:
Hi @rmsyn ! I left you a few comments. Let's first figure out how we want to use the macros and then we can explore using the `paste` crate (which looks great, by the way).

rmsyn:
> Looking at this specific example, the macro calls other macros (`csr`, `read_csr_as`, and `read_only_csr_field` a few times). So what is the benefit of having a single macro with many pattern branches that just propagate to other macro calls? Why don't we _keep it simple_ and call in the user code to the `read_only_csr_field` all the times we want? In this way, we keep all macros simpler and users call them as many times as required.

Ah, I see what you mean now. Yes, the code could be simplified by not trying to include field definitions in the type-creation macro.

The main instance where including field definitions in the type-creation macro is when all fields are the same type (bit, range, field-mask, etc.). Otherwise, users will need to invoke the field macros anyway.

I agree simplifying is the way to go here.

riscv/src/register/macros.rs
```diff
@@ -600,3 +600,1024 @@ macro_rules! clear_pmp {
...
+            /// Creates a new CSR type from raw bits value.
+            ///
+            /// Only bits in the [BITMASK](Self::BITMASK) will be set.
+            pub const fn from_bits(bits: usize) -> Self {
+                Self { bits: bits & $mask }
+            }
+
+            /// Gets the raw bits value.
+            pub const fn bits(&self) -> usize {
+                self.bits & $mask
```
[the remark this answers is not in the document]
rmsyn: I think it's still important to apply the mask for when the register is read from the CSR, and not using the `from_bits` constructor.

Is there a benefit to removing the mask?

riscv/src/register/macros.rs
```diff
@@ -600,3 +600,1024 @@ macro_rules! clear_pmp {
...
+         }
+
+         impl $field_ty {
+             /// Creates a new field variant.
+             pub const fn new() -> Self {
+                 Self::$default_variant
+             }
+
+             /// Attempts to convert a [`usize`] into a valid variant.
+             pub const fn from_usize(val: usize) -> Option<Self> {
```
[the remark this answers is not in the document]
rmsyn: Yes, I missed this when converting after the introduction of the `Result` type.

riscv/src/register/macros.rs
```diff
@@ -600,3 +600,1024 @@ macro_rules! clear_pmp {
...
+             fn default() -> Self {
+                 Self::new()
+             }
+         }
+
+         impl From<$field_ty> for usize {
+             fn from(val: $field_ty) -> Self {
+                 val.into_usize()
+             }
+         }
```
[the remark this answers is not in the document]
rmsyn: The `TryFrom` is automatically implemented by the `From` implementation.

riscv/src/register/macros.rs
rmsyn: When working on this, I was considering the safety of the `bits` functions. There should be `try_` variants of the `bf_insert` and `bf_extract` functions, because both of these functions can implicitly panic.

We would use the `checked_shl` function instead of `<<` operator.

What would the compile-time check look like here, as an example?

rmsyn:
> Let's first figure out how we want to use the macros and then we can explore using the paste crate (which looks great, by the way).

I've moved the `paste` work into [riscv/csr-macro-paste](https://github.com/rmsyn/riscv/tree/riscv/csr-macro-paste) branch.

When we're settled on the other design work, I can bring that work into this branch.

riscv/src/register/macros.rs
```diff
@@ -397,3 +397,681 @@ macro_rules! clear_pmp {
...
+                $field: $get_bit,
+                $set_field_doc,
+                $set_field: $set_bit,
+            );
+        )+
+
+        $crate::read_csr_as!($ty, $csr);
+        $crate::write_csr_as!($ty, $csr);
+        $crate::set!($csr);
+        $crate::clear!($csr);
```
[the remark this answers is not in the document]
rmsyn: I agree about leaving out the `set` and `clear` macros, but think the `read_csr_as` and `write_csr_as` shoudl still be included.

Would there ever be a case where one would call:

- `read_write_csr`, but not `read_csr_as` and `write_csr_as`
- `read_only_csr`, but not `read_csr_as`
- `write_only_csr`, but not `write_csr_as`

In my opinion, including these macro calls avoids an author forgetting to include them. If you really think they should be removed, I'll take them out.

riscv/src/register/macros.rs
rmsyn: Similarly as above, can you provide an example of what the compile-time check would look like?

riscv/src/register/macros.rs
romancardenas: The `$bit` param is fixed, right? For example, the bit number 3 of a register. Thus, we can add something like [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=cd70233e4269326d495c04dddda11f63) to check at compile time that `$bit` is within the expected bounds. In this way, we don't need `try_` methods

riscv/src/register/macros.rs
romancardenas: Same as above. We add compile-time checks for `start` and `end`. Then, we only need to check that the parameter index is within the expected range.

riscv/src/register/macros.rs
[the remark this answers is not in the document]
romancardenas: I mean `TryFrom<usize> for $field_ty`

riscv/src/register/macros.rs
```diff
@@ -600,3 +600,510 @@ macro_rules! clear_pmp {
...
+            }
+        }
+    };
+
+    ($ty:ident,
+     $(#[$field_doc:meta])+
+     $field:ident,
+     $(#[$try_field_doc:meta])+
+     $try_field:ident,
+     range: $bit_start:literal..=$bit_end:literal$(,)?) => {
```
romancardenas, line 882: I would use to `start : end` for consistency with the other branches
```suggestion
     range: $bit_start:literal : $bit_end:literal$(,)?) => {
```

riscv/src/register/macros.rs
```diff
@@ -600,3 +600,510 @@ macro_rules! clear_pmp {
...
+        const _: () = assert!($bit_end < usize::BITS);
+        const _: () = assert!($bit_start <= $bit_end);
```
romancardenas, line 884: Why not only `<`? to ensure that there are at least two bits

```suggestion
        const _: () = assert!($bit_start < $bit_end);
```

riscv/src/register/macros.rs
```diff
@@ -600,3 +600,510 @@ macro_rules! clear_pmp {
...
+        $crate::read_only_csr_field!(
+            $ty,
+            $(#[$field_doc])*
+            $field,
+            $(#[$try_field_doc])*
+            $try_field,
+            $field_ty,
+            range: [$field_start : $field_end],
+        );
+    };
```
romancardenas, line 954: I would remove this branch and define the enums and the methods separately.

```suggestion
```

riscv/src/register/macros.rs
```diff
@@ -600,3 +600,510 @@ macro_rules! clear_pmp {
...
+            }
+        }
+    };
+
+    ($ty:ident,
+     $(#[$field_doc:meta])+
+     $field:ident,
+     $(#[$try_field_doc:meta])+
+     $try_field:ident,
+     range: $bit_start:literal..=$bit_end:literal$(,)?) => {
```
romancardenas, line 1011: Same here:

```suggestion
     range: $bit_start:literal: $bit_end:literal$(,)?) => {
```

riscv/src/register/macros.rs
```diff
@@ -600,3 +600,510 @@ macro_rules! clear_pmp {
...
+        const _: () = assert!($bit_end < usize::BITS);
+        const _: () = assert!($bit_start <= $bit_end);
```
romancardenas, line 1013: Same here:

```suggestion
        const _: () = assert!($bit_start < $bit_end);
```

riscv/src/register/macros.rs
```diff
@@ -600,3 +600,510 @@ macro_rules! clear_pmp {
...
+         );
+
+        $crate::write_only_csr_field!(
+            $ty,
+            $(#[$field_doc])+
+            $field,
+            $field_ty,
+            range: [$field_start : $field_end],
+        );
+    };
```
romancardenas, line 1085: Same here:

```suggestion
```

romancardenas (commented):
I have been without an Internet connection, but I had some time to review this PR. Check my comments.

riscv/src/register/macros.rs
rmsyn: This is specifically to handle a range of single-bit fields, like `hpm` in `mcountinhibit`.

The `[start : end]` variant is for a multi-bit field.

riscv/src/register/macros.rs
rmsyn: I included the `<=` because it is technically valid, even if it probably would only ever be a typo, e.g. user's would probably want the single-bit variant if they were defining a field where `start == end`.

I can change it to `<`.

rmsyn: I included the `<=` because it is technically valid, even if it probably would only ever be a typo, e.g. users would probably want the single-bit variant if they were defining a field where `start == end`.

I can change it to `<`.

riscv/src/register/macros.rs
rmsyn: Same as above, this branch is to define a range of single-bit fields.

riscv/src/register/macros.rs
```diff
@@ -600,3 +600,445 @@ macro_rules! clear_pmp {
...
+    ($ty:ident,
+     $(#[$field_doc:meta])+
+     $field:ident,
+     $(#[$try_field_doc:meta])+
+     $try_field:ident,
+     $field_ty:ident,
+     range: [$field_start:literal : $field_end:literal]$(,)?
+    ) => {
+        const _: () = assert!($field_end < usize::BITS);
+        const _: () = assert!($field_start <= $field_end);
```
rmsyn, line 928: For the enum branches, I left the `<=` check.

My thinking here is that there may be CSR definitions where users want to use an enum to clarify the meaning of a single-bit field.

For example, a platform with a custom CSR where a single-bit field can mean:

- `0`: set tolerance to custom value X
- `1`: set tolerance to custom value Y

And the enum definition would be something like:

```rust
csr_field_enum! {
   /// Some custom CSR field.
   CsrFieldTolerance,
   range: [0:0],
   default: ToleranceX,
   ToleranceX = 0,
   ToleranceY = 1,
}
```

riscv/src/register/tests/read_only_csr.rs
```diff
@@ -0,0 +1,127 @@
...
+
+#[test]
+fn test_mtest_read_only() {
+    let mut mtest = Mtest::from_bits(0);
+
+    assert_eq!(mtest.bitmask(), Mtest::BITMASK);
+    assert_eq!(mtest.bits(), 0);
+
+    // check that single bit field getter/setters work.
+    assert_eq!(mtest.single(), false);
```
romancardenas, line 71: Clippy complains at my machine about these kinds of assertions. Should look like:

```suggestion
    assert!(!mtest.single());
```

romancardenas (commented):
Ok so... Looks almost there!

What do you think about trying the `paste` thing so we can shrink even more the macros? The idea would be:

- For getters of field `$field`: `$field` (I think there is no fallible getter?)
- For setters: `try_$field` (fallible) and `set_$field` (infallible)

If you prefer, we can address this in another PR and move on.

rmsyn:
> Looks like Clippy nightly now complains about [this](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/static-mut-references.html).

Oh OK, I was wondering why some of the lint checks were failing in unrelated code. Makes sense now.

> We should take a look, perhaps using `static UnsafeCell`. Another option is just deprecating `riscv-semihosting` and tell users to use [`semihosting`](https://github.com/taiki-e/semihosting), which is way more advanced.

I'm not sure on that, and would need to take some time to weigh the trade-offs.

At first glance, the fact that [`semihosting`](https://github.com/taiki-e/semihosting) supports multiple architectures might be better for users supporting multiple platforms in a single crate. No strong opinions either way.

riscv/Cargo.toml
```diff
@@ -27,3 +27,4 @@ critical-section-single-hart = ["critical-section/restore-state-bool"]
 critical-section = "1.1.2"
 embedded-hal = "1.0.0"
 riscv-pac = { path = "../riscv-pac", version = "0.1.1", default-features = false }
+paste = "1.0.15"
```
romancardenas, line 30: Just to follow the alphabet:
```suggestion
paste = "1.0.15"
riscv-pac = { path = "../riscv-pac", version = "0.1.1", default-features = false }
```

romancardenas (commented):
LGTM! Just a minor, silly comment regarding the order of the dependencies in Cargo.toml.

Also, the `paste` dependency is considerably lightweight, I don't see any issue adding it.

If we are OK with `paste`, we can further simplify some macros (e.g., `set_clear_csr`). But again, I'd leave these potential changes for another PR, this one looks big enough to me.

rmsyn:
> If we are OK with paste, we can further simplify some macros (e.g., set_clear_csr). But again, I'd leave these potential changes for another PR, this one looks big enough to me.

Definitely, `paste` opens up a lot of opportunities for simplifying / reducing macros. Leaving it for another PR also makes sense.

riscv/src/register/macros.rs
```diff
@@ -600,3 +600,431 @@ macro_rules! clear_pmp {
...
+
+    ($ty:ident,
+     $(#[$field_doc:meta])+
+     $field:ident,
+     $field_ty:ident: [$field_start:literal : $field_end:literal]$(,)?
+    ) => {
+        const _: () = assert!($field_end < usize::BITS);
+        const _: () = assert!($field_start <= $field_end);
+
+        paste::paste! {
```
romancardenas, line 906: shouldn't we re-export `paste` to allow external users to use this macros? something like `$crate::paste!`

romancardenas (commented):
I noticed that we may need to re-export `paste`. @rmsyn , can you confirm this? If not, I think this PR is ready.

riscv/src/register/macros.rs
rmsyn: Good catch, I added a re-export commit.

romancardenas (approved):
LGTM

Let's leave a couple of days in case someone wants to review this and merge it

romancardenas:
If nobody opposes, I will merge this PR tomorrow by the end of the weekly meeting

romancardenas:
@rmsyn can you deal with the conflicts? We will merge it to master right after fixing this.

romancardenas (approved):

PR #220, opened 2024-06-07
mstatus: Support vector extension

jasonwhite:
Adds the ability to turn on the vector extension.

### DIFF
--- a/riscv/src/register/mstatus.rs
+++ b/riscv/src/register/mstatus.rs
@@ -41,6 +41,15 @@ pub enum FS {
     Dirty = 3,
 }
 
+/// Vector extension state
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum VS {
+    Off = 0,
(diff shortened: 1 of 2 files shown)


riscv/src/register/mstatus.rs
```diff
@@ -41,6 +41,15 @@ pub enum FS {
     Dirty = 3,
 }
 
+/// Vector extension state
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum VS {
+    Off = 0,
+    Initial = 1,
+    Clean = 2,
+    Dirty = 3,
```
jasonwhite, line 50: Unsure if we want to reuse the `XS` enum instead of creating a brand new enum. However, I'm just following the pattern of `FS` here.

romancardenas (approved):
LGTM.

In the near future, we might want to unify the data type and use the `XS` enum instead of having custom enums for every potential extension.

PR #221, opened 2024-06-23
riscv-rt-macros: update syn to v2.0

agarof:
Upgrade `syn` in order to deduplicate dependencies in projects depending on `riscv-rt`, as most other crates have upgraded to version 2.
Tested by building and flashing a simple hello world to the ESP32-C3, with no difference between the two versions.

While syn's [README](https://github.com/dtolnay/syn/blob/master/README.md) states `Version requirement: Syn supports rustc 1.61 and up.`, riscv-rt-macros still builds with `1.59`.

### DIFF
--- a/.github/workflows/riscv-rt.yaml
+++ b/.github/workflows/riscv-rt.yaml
@@ -10,8 +10,8 @@ jobs:
   build:
     strategy:
       matrix:
-        # All generated code should be running on stable now, MRSV is 1.60.0
-        toolchain: [ stable, nightly, 1.60.0 ]
+        # All generated code should be running on stable now, MRSV is 1.61.0
+        toolchain: [ stable, nightly, 1.61.0 ]
         target:
           - riscv32i-unknown-none-elf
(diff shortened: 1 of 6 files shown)


romancardenas:
Thanks for the PR! It looks like we need to bump the MSRV to 1.61

agarof:
I bumped the MSRV of `riscv-rt` to 1.61, CI should pass now

almindor (approved):

almindor:
Thank you!

almindor:
I think this might be good as is nevertheless we should also bump the CI rust versions at https://github.com/rust-embedded/riscv/blob/master/.github/workflows/riscv-rt.yaml#L13-L14

agarof:
The change has already been done here : https://github.com/rust-embedded/riscv/pull/221/files#diff-0ec8e1c75aebc15aec4ac8becbc3c0aa0e7980346b90cf55ba03dedde8630911R14

romancardenas (approved):

PR #222, opened 2024-06-28
riscv: add fallible functions

rmsyn:
Adds fallible variants to all functions that explicitly panic in the `riscv` crate.

Resolves: https://github.com/rust-embedded/riscv/issues/212

### DIFF
--- a/riscv-pac/src/result.rs
+++ b/riscv-pac/src/result.rs
@@ -13,13 +13,17 @@ pub enum Error {
         max: usize,
     },
     /// Invalid field value.
-    InvalidValue {
+    InvalidFieldValue {
         field: &'static str,
         value: usize,
         bitmask: usize,
     },
     /// Invalid value of a register field that does not match any known variants.
-    InvalidVariant { field: &'static str, value: usize },
+    InvalidFieldVariant { field: &'static str, value: usize },
+    /// Invalid value.
+    InvalidValue { value: usize, bitmask: usize },
+    /// Invalid value that does not match any known variants.
+    InvalidVariant(usize),
     /// Unimplemented function or type.
     Unimplemented,
 }
@@ -31,7 +35,7 @@ impl fmt::Display for Error {
                 f,
                 "out-of-bounds access, index: {index}, min: {min}, max: {max}"
             ),
-            Self::InvalidValue {
+            Self::InvalidFieldValue {
                 field,
                 value,
                 bitmask,
@@ -39,8 +43,14 @@ impl fmt::Display for Error {
                 f,
                 "invalid {field} field value: {value:#x}, valid bitmask: {bitmask:#x}",
             ),
-            Self::InvalidVariant { field, value } => {
+            Self::InvalidFieldVariant { field, value } => {
                 write!(f, "invalid {field} field variant: {value:#x}")
+            }
+            Self::InvalidValue { value, bitmask } => {
+                write!(f, "invalid value: {value:#x}, valid bitmask: {bitmask:#x}",)
+            }
+            Self::InvalidVariant(value) => {
+                write!(f, "invalid variant: {value:#x}")
             }
             Self::Unimplemented => write!(f, "unimplemented"),
         }
--- a/riscv-pac/src/result.rs
+++ b/riscv-pac/src/result.rs
@@ -1,0 +1,48 @@
+use core::fmt;
+
+/// Convenience alias for the [Result](core::result::Result) type for the library.
+pub type Result<T> = core::result::Result<T, Error>;
+
+/// Represents error variants for the library.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum Error {
+    /// Attempted out-of-bounds access.
+    IndexOutOfBounds {
+        index: usize,
+        min: usize,
+        max: usize,
+    },
+    /// Invalid field value.
+    InvalidValue {
+        field: &'static str,
+        value: usize,
+        bitmask: usize,
+    },
+    /// Invalid value of a register field that does not match any known variants.
+    InvalidVariant { field: &'static str, value: usize },
+    /// Unimplemented function or type.
+    Unimplemented,
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::IndexOutOfBounds { index, min, max } => write!(
+                f,
+                "out-of-bounds access, index: {index}, min: {min}, max: {max}"
+            ),
+            Self::InvalidValue {
+                field,
+                value,
+                bitmask,
+            } => write!(
+                f,
+                "invalid {field} field value: {value:#x}, valid bitmask: {bitmask:#x}",
+            ),
+            Self::InvalidVariant { field, value } => {
+                write!(f, "invalid {field} field variant: {value:#x}")
+            }
+            Self::Unimplemented => write!(f, "unimplemented"),
+        }
+    }
+}
--- a/riscv-peripheral/examples/e310x.rs
+++ b/riscv-peripheral/examples/e310x.rs
@@ -2,6 +2,7 @@
 //! This is a simple example of how to use the `riscv-peripheral` crate to generate
 //! peripheral definitions for a target.
 
+use riscv_pac::result::{Error, Result};
 use riscv_pac::{HartIdNumber, InterruptNumber, PriorityNumber};
 
 #[repr(u16)]
@@ -19,9 +20,12 @@ unsafe impl HartIdNumber for HartId {
     }
 
     #[inline]
-    fn from_number(number: u16) -> Result<Self, u16> {
+    fn from_number(number: u16) -> Result<Self> {
         if number > Self::MAX_HART_ID_NUMBER {
-            Err(number)
+            Err(Error::InvalidVariant {
+                field: "hart_id",
+                value: number as usize,
+            })
         } else {
             // SAFETY: valid context number
             Ok(unsafe { core::mem::transmute(number) })
@@ -95,9 +99,12 @@ unsafe impl InterruptNumber for Interrupt {
     }
 
     #[inline]
-    fn from_number(number: u16) -> Result<Self, u16> {
+    fn from_number(number: u16) -> Result<Self> {
         if number == 0 || number > Self::MAX_INTERRUPT_NUMBER {
-            Err(number)
+            Err(Error::InvalidVariant {
+                field: "interrupt",
+                value: number as usize,
+            })
         } else {
             // SAFETY: valid interrupt number
             Ok(unsafe { core::mem::transmute(number) })
@@ -127,9 +134,12 @@ unsafe impl PriorityNumber for Priority {
     }
 
     #[inline]
-    fn from_number(number: u8) -> Result<Self, u8> {
+    fn from_number(number: u8) -> Result<Self> {
         if number > Self::MAX_PRIORITY_NUMBER {
-            Err(number)
+            Err(Error::InvalidVariant {
+                field: "priority",
+                value: number as usize,
+            })
         } else {
             // SAFETY: valid priority number
             Ok(unsafe { core::mem::transmute(number) })
--- a/riscv-peripheral/examples/e310x.rs
+++ b/riscv-peripheral/examples/e310x.rs
@@ -22,10 +22,7 @@ unsafe impl HartIdNumber for HartId {
     #[inline]
     fn from_number(number: u16) -> Result<Self> {
         if number > Self::MAX_HART_ID_NUMBER {
-            Err(Error::InvalidVariant {
-                field: "hart_id",
-                value: number as usize,
-            })
+            Err(Error::InvalidVariant(number as usize))
         } else {
             // SAFETY: valid context number
             Ok(unsafe { core::mem::transmute(number) })
@@ -101,10 +98,7 @@ unsafe impl InterruptNumber for Interrupt {
     #[inline]
     fn from_number(number: u16) -> Result<Self> {
         if number == 0 || number > Self::MAX_INTERRUPT_NUMBER {
-            Err(Error::InvalidVariant {
-                field: "interrupt",
-                value: number as usize,
-            })
+            Err(Error::InvalidVariant(number as usize))
         } else {
             // SAFETY: valid interrupt number
             Ok(unsafe { core::mem::transmute(number) })
@@ -136,10 +130,7 @@ unsafe impl PriorityNumber for Priority {
     #[inline]
     fn from_number(number: u8) -> Result<Self> {
         if number > Self::MAX_PRIORITY_NUMBER {
-            Err(Error::InvalidVariant {
-                field: "priority",
-                value: number as usize,
-            })
+            Err(Error::InvalidVariant(number as usize))
         } else {
             // SAFETY: valid priority number
             Ok(unsafe { core::mem::transmute(number) })
--- a/riscv/src/lib.rs
+++ b/riscv/src/lib.rs
@@ -40,6 +40,7 @@ pub(crate) mod bits;
 pub mod delay;
 pub mod interrupt;
 pub mod register;
+pub use riscv_pac::*;
 
 #[macro_use]
 mod macros;
--- a/riscv/src/register/macros.rs
+++ b/riscv/src/register/macros.rs
@@ -6,19 +6,27 @@
 #[macro_export]
 macro_rules! read_csr {
     ($csr_number:literal) => {
-        /// Reads the CSR
+        /// Reads the CSR.
+        ///
+        /// **WARNING**: panics on non-`riscv` targets.
         #[inline]
         unsafe fn _read() -> usize {
+            _try_read().unwrap()
+        }
+
+        /// Attempts to read the CSR.
+        #[inline]
+        unsafe fn _try_read() -> $crate::result::Result<usize> {
             match () {
                 #[cfg(any(target_arch = "riscv32", target_arch = "riscv64"))]
                 () => {
                     let r: usize;
                     core::arch::asm!(concat!("csrrs {0}, ", stringify!($csr_number), ", x0"), out(reg) r);
-                    r
+                    Ok(r)
                 }
 
                 #[cfg(not(any(target_arch = "riscv32", target_arch = "riscv64")))]
-                () => unimplemented!(),
+                () => Err($crate::result::Error::Unimplemented),
             }
         }
     };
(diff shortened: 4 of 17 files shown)


rmsyn:
We probably also want to cover the `riscv::register::satp` and `riscv::register::pmpcfgx` modules, too.

Wanted to submit the initial work to get feedback on the approach.

riscv/src/register/macros.rs
```diff
@@ -65,6 +101,16 @@ macro_rules! read_csr_as {
                 bits: unsafe { _read() },
             }
         }
+
+        /// Attempts to reads the CSR.
+        ///
+        /// Only implemented for `riscv32` and `riscv64` targets.
```
romancardenas, line 107: We can remove this

romancardenas (commented):
Looks great! Thank you so much. We can keep working in this PR

riscv/src/register/satp.rs
```diff
@@ -114,3 +138,25 @@ pub unsafe fn set(mode: Mode, asid: usize, ppn: usize) {
     let bits = (mode as usize) << 60 | (asid << 44) | ppn;
     _write(bits);
 }
+
+/// Sets the register to corresponding page table mode, physical page number and address space id.
```
rmsyn, line 142:
```suggestion
/// Attempts to set the register to corresponding page table mode, physical page number and address space id.
```

Missed the update to the docs, will add with next set of changes.

riscv/src/register/pmpcfgx.rs
```diff
@@ -13,6 +15,28 @@ pub enum Permission {
...
+    fn try_from(val: u8) -> Result<Self> {
+        match val {
+            0b000 => Ok(Self::NONE),
+            0b001 => Ok(Self::R),
+            0b010 => Ok(Self::W),
+            0b011 => Ok(Self::RW),
+            0b100 => Ok(Self::X),
+            0b101 => Ok(Self::RX),
+            0b110 => Ok(Self::WX),
+            0b111 => Ok(Self::RWX),
```
romancardenas, line 30: what if valid values go from 0 to 6? I suggest using two different errors: `InvalidValue` (more generic) and `OutOfBounds` (taking into account the mask stuff). We can even define a different enum for typical `InvalidValue` types (`OutOfBounds`, `Unknown` or something like that)

riscv/src/register/pmpcfgx.rs
```diff
@@ -76,6 +118,37 @@ impl Pmpcsr {
...
+    /// Attempts to take the register contents, and translate into a Pmp configuration struct.
+    #[inline]
+    pub fn try_into_config(&self, index: usize) -> Result<Pmp> {
+        let max = if cfg!(riscv32) {
+            Ok(4usize)
+        } else if cfg!(riscv64) {
+            Ok(8usize)
+        } else {
+            Err(Error::Unimplemented)
+        }?;
```
romancardenas, line 131:
```suggestion
        let max = if cfg!(riscv32) {
            4usize
        } else if cfg!(riscv64) {
            8usize
        } else {
            return Err(Error::Unimplemented);
        };
```

riscv/src/result.rs
```diff
@@ -0,0 +1,44 @@
+use core::fmt;
+
+/// Convenience alias for the [Result](core::result::Result) type for the library.
+pub type Result<T> = core::result::Result<T, Error>;
+
+/// Represents error variants for the library.
+#[repr(C)]
```
romancardenas, line 7: Just curious: how does `#[repr(C)]` work with complex enums?

riscv/src/result.rs
```diff
@@ -0,0 +1,44 @@
...
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum Error {
+    /// Attempted out-of-bounds access.
+    OutOfBounds {
```
romancardenas, line 11:
```suggestion
    IndexOutOfBounds {
```

riscv/src/result.rs
```diff
@@ -0,0 +1,44 @@
...
+        min: usize,
+        max: usize,
+    },
+    /// Unimplemented function or type.
+    Unimplemented,
+    /// Invalid field value.
+    InvalidValue {
+        field: &'static str,
+        value: usize,
+        bitmask: usize,
```
romancardenas, line 22: Maybe we can use something like:

```rust
pub enum InvalidValue {
    OutOfBounds {
        bitmask: usize,
    },
    ... // other causes
}
```

Or maybe this is overcomplicating stuff...

romancardenas (commented):
It looks good to me. What do you think about more details in invalid values? Is it overcomplicating stuff maybe?

riscv/src/register/macros.rs
```diff
@@ -377,6 +557,39 @@ macro_rules! set_pmp {
...
+            permission: Permission,
+            locked: bool,
+        ) -> $crate::result::Result<()> {
+            let max = if cfg!(target_arch = "riscv32") {
+                Ok(4usize)
+            } else if cfg!(target_arch = "riscv64") {
+                Ok(8usize)
+            } else {
+                Err($crate::result::Error::Unimplemented)
+            }?;
```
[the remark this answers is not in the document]
rmsyn: Something like:

```rust
let max = match () {
    () => Ok(4usize),
    #[cfg(target_arch = "riscv64")]
    () => Ok(8usize),
    #[cfg(not(any(target_arch = "riscv32", target_arch = "riscv64")))]
    () => Err($crate::result::Error::Unimplemented),
}?;
```

I can try that.

riscv/src/register/macros.rs
```diff
@@ -395,5 +608,31 @@ macro_rules! clear_pmp {
...
+        /// Returns an error if the index is invalid.
+        #[inline]
+        pub unsafe fn try_clear_pmp(index: usize) -> $crate::result::Result<()> {
+            let max = if cfg!(target_arch = "riscv32") {
+                Ok(4usize)
+            } else if cfg!(target_arch = "riscv64") {
+                Ok(8usize)
+            } else {
+                Err($crate::result::Error::Unimplemented)
+            }?;
```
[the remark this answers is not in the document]
rmsyn: Kind of the same either way, right? Down to whether one prefers the explicit `return`, or the `?` operator.

riscv/src/register/pmpcfgx.rs
rmsyn: Not sure what you want here.

Currently, the `Error::InvalidValue` variant is used to signify an invalid bitfield value for a register.

`Error::OutOfBounds` is used to indicate an invalid index access outside an acceptable range.

Are you suggesting the the `Error::InvalidValue` variant itself be a wrapper around a `enum InvalidValue` type? I think this complicates things too much, if that's what you're suggesting.

riscv/src/register/pmpcfgx.rs
rmsyn: Is there a reason you prefer the explicit `return` keyword?

From what I've seen in style guides, and personal practice, having one `return` statement max is preferred. Every early return should utilize the `?` operator where possible.

riscv/src/result.rs
rmsyn: Somewhat of a habit to use `#[repr(C)]` on types that could potentially cross an FFI boundary, though I'm not exactly sure how Rust handles it for complex enums.

From [godbolt](https://godbolt.org/z/hnW7q8xj7) it looks like it compiles to a flat structure with pointer offsets.

I can remove it if you want.

riscv/src/result.rs
rmsyn: > Or maybe this is overcomplicating stuff...

I think so, the print out of the error tells the reader that an invalid field value exists, what the value is, and what the valid bitmask is.

Easy enough for the person debugging the error to convert the values to binary, and see what bits don't match up.

Also what would semantically differentiate an "OutOfBounds" from an "Unknown" (or other variant). They would all technically fall outside the acceptable bitmask, no?

Or do you mean "OutOfBounds" as "value with a bit set below minimum, and/or value with a bit set above maximum" and "Unknown" as a "value with invalid bits set within the bitmask, e.g. bitmask = 0b101, value = 0b111"?

Either way, I think a user can get basically the same information from what exists now. We may want to format the output as binary instead of the current hex, though.

rmsyn:
I've added suggested fixes as `fixup` commits for easier review.

If you're happy with them, I'll squash them into their respective parent commits before a merge.

I've also left conversations open to continue discussion on some of the style-related suggestions.

riscv/src/result.rs
romancardenas: I'm not an expert in these issues, but I would say that results are quite rustacean, and error types are not usually exposed in FFIs. Instead, we would expose the "infallible" functions.

romancardenas (commented):
It is looking great! The only comments I have are:

- Adding an `InvalidVariant` error for those cases in which there are reserved values within a bitmask.
- Updating "infallible" functions to use their "fallible" counterpart followed by an `unwrap()`. In this way, new panic messages will show a more comprehensive outcome.

rmsyn:
Ok, I added the next round of `fixup` changes.

riscv/src/register/mcounteren.rs
```diff
@@ -52,19 +53,62 @@ impl Mcounteren {
     }
 
     /// Supervisor "hpm\[x\]" Enable (bits 3-31)
+    ///
+    /// **WARNING**: panics on:
+    ///
+    /// - non-`riscv` targets
+    /// - `index` out-of-bounds
```
romancardenas, line 60:
```suggestion
    /// **WARNING**: panics on `index` out-of-bounds
```

riscv/src/register/mcounteren.rs
```diff
@@ -52,19 +53,62 @@ impl Mcounteren {
...
     }
 
     /// Sets whether to enable the "hpm\[X\]" counter.
     ///
     /// Only updates the in-memory value, does not modify the `mcounteren` register.
+    ///
+    /// **WARNING**: panics on:
+    ///
+    /// - non-`riscv` targets
+    /// - `index` out-of-bounds
```
romancardenas, line 89:
```suggestion
    /// **WARNING**: panics on `index` out-of-bounds
```

riscv/src/register/satp.rs
```diff
@@ -14,27 +16,35 @@ impl Satp {
...
-        }
+        self.try_mode().unwrap()
+    }
+
+    /// Attempts to get the current address-translation scheme.
+    #[inline]
+    #[cfg(target_pointer_width = "32")]
+    pub fn try_mode(&self) -> Result<Mode> {
+        ((self.bits >> 31) as u8).try_into()
     }
```
romancardenas, line 32: Not sure if it is worth it to have the `try_mode` and `mode` methods. Theoretically, the struct should never hold invalid values, right? Maybe only the `mode` method with the `unwrap()` is enough.

romancardenas (commented):
So far so good!

I was thinking about the correct location of the `result` module and maybe we should consider moving it to `riscv-pac`. Ideally, `riscv-pac` will have way less activity and act as a dependency for all the `riscv` ecosystem. Moving it to `riscv-pac` will allow us to bump `riscv` versions without breaking other crates, as there will not be version issues.

riscv/src/register/satp.rs
rmsyn: I wasn't sure, either. For consistency with the `target_pointer_width = "64"` implementation, I included it.

I don't think it really harms anything to have the `try_mode` function for the `target_pointer_width = "32"` implementation.

It would also allow for less feature-gating downstream to keep `try_mode` for both implementations.

rmsyn:
> I was thinking about the correct location of the result module and maybe we should consider moving it to riscv-pac. Ideally, riscv-pac will have way less activity and act as a dependency for all the riscv ecosystem.

I'm not sure about the "correct" location for the `result` module. Putting it in `riscv-pac` would require adding `riscv-pac` as a `riscv` dependency, where keeping it in `riscv` requires the opposite.

I think the latter makes more sense, and follows with `riscv-rt` having a `riscv` dependency.

> Moving it to riscv-pac will allow us to bump riscv versions without breaking other crates, as there will not be version issues.

We could still bump `riscv` versions, we would just need to do `semver` bumps for any `rust-embedded/riscv` workspace crates that have a `riscv` dependency.

Maybe I'm misunderstanding the architecture, but I thought `riscv` was the "core" crate that all the `riscv-*` crates would bring in as a dependency (if they need it).

romancardenas:
You can get more information about the motivation of the `riscv-pac` crate [here](https://github.com/rust-embedded/cortex-m/pull/488).


You can think of `riscv-pac` as a "special" module of `riscv` that is **completely** isolated from the rest of the crate. In future implementations, I foresee `riscv` doing a public reexport of `riscv-pac`. `riscv-pac` will experience little to no breaking changes, while we can keep improving `riscv` without too big impacts in the ecosystem.

Let us assume that we have the `ExternalInterruptNumber` trait in `riscv` instead of `riscv-pac`. The `riscv_peripheral::PLIC` peripheral implementation relies on this trait to provide safe access to different registers. If we add breaking changes to `riscv`, we then should also release a new version of `riscv-peripheral`, even if no changes were made in the `ExternalInterruptNumber` trait. This scales to PACs, third-party peripheral drivers, and so on.

If we place the `result` module in `riscv-pac`, it will be available in `riscv` with the reexport. Additionally, we can improve the current `riscv-pac` traits by using your new `Result` enum.

rmsyn:
> If we place the result module in riscv-pac, it will be available in riscv with the reexport. Additionally, we can improve the current riscv-pac traits by using your new Result enum.

I think this makes sense. If you are planning to have `riscv-pac` be a dependency in `riscv` anyway, then moving `result` into `riscv-pac` doesn't cause any issues.

In fact, if that's the plan, then moving `result` into `riscv-pac` is a requirement to avoid a circular dependency.

A similar approach would still need to be taken for any additions to the `Error` enum, where breaking changes would require a `semver` bump in `riscv-pac`, and everything that re-exports `riscv_pac::result`.

I'll move it in a `fixup` commit, and if you like it, squash the commit before merge.

riscv/src/lib.rs
```diff
@@ -40,6 +40,7 @@ pub(crate) mod bits;
 pub mod delay;
 pub mod interrupt;
 pub mod register;
+pub use riscv_pac::result;
```
romancardenas, line 43: It could be handy to reexport the `riscv-pac` crate
```suggestion
pub use riscv_pac::{self as pac, result};
```

romancardenas (commented):
LGTM

riscv/src/lib.rs
romancardenas: ... or even `pub use riscv_pac::*`?

romancardenas (commented):
LGTM

romancardenas:
I think this is done. The main new features are:

- New `result` module in `riscv-pac`, in which we define the typical error cases for embedded RISC-V.
- Now, in `riscv`, we do `pub use riscv_pac::*`. Thus, `riscv-pac` could be considered just an "extension" of `riscv` that isolates traits and error enums from other breaking changes. Also, in this way, PACs would not need to add an extra dependency to `riscv-pac`.

While there are other aspects to polish (e.g., modifying `riscv-pac` traits so they use the new `Error`s), I would like to wait for the next WG meeting (we are already on the [agenda](https://github.com/rust-embedded/wg/discussions/773#discussioncomment-9940370)). The idea is to get some consensus so the `cortex-m` ecosystem could replicate this work. Also, there are very knowledgeable folks who have been working on `embedded-hal` and all sorts of interfaces that could help us improve the user experience and avoid potential issues in the future.

romancardenas (commented):
LGTM. Now we need to update `riscv-peripheral` 😁

rmsyn:
> LGTM. Now we need to update `riscv-peripheral` 😁

:sweat_smile: I looked for current usage of the traits, and must have missed `riscv-peripheral`. Updating now.

riscv-pac/src/result.rs
```diff
@@ -0,0 +1,48 @@
...
+        max: usize,
+    },
+    /// Invalid field value.
+    InvalidValue {
+        field: &'static str,
+        value: usize,
+        bitmask: usize,
+    },
+    /// Invalid value of a register field that does not match any known variants.
+    InvalidVariant { field: &'static str, value: usize },
```
romancardenas, line 22: Maybe we could differentiate errors regarding fields of a register and errors regarding the whole register/value. The idea is avoiding defining the `field` in the `riscv-peripheral` examples, as I think it is not necessary.

```suggestion
    /// Invalid value of a register field.
    InvalidFieldValue {
        field: &'static str,
        value: usize,
        bitmask: usize,
    },
    /// Invalid value of a register field that does not match any known variants.
    InvalidFieldVariant { field: &'static str, value: usize },
     /// Invalid value.
     InvalidValue {
        value: usize,
        bitmask: usize,
    },
    /// Invalid value that does not match any known variants.
    InvalidVariant(usize),
```

riscv-peripheral/examples/e310x.rs
```diff
@@ -19,9 +20,12 @@ unsafe impl HartIdNumber for HartId {
     }
 
     #[inline]
-    fn from_number(number: u16) -> Result<Self, u16> {
+    fn from_number(number: u16) -> Result<Self> {
         if number > Self::MAX_HART_ID_NUMBER {
-            Err(number)
+            Err(Error::InvalidVariant {
+                field: "hart_id",
```
romancardenas, line 26: Adding `field` to these errors feels unnecessary to me.

romancardenas (commented):
Let me know what you think about my cmoments

riscv-pac/src/result.rs
rmsyn: > Maybe we could differentiate errors regarding fields of a register and errors regarding the whole register/value.

This makes sense to me. The value covers the entire structure, so this separation could be useful in other cases without subfields.

rmsyn:
> Let me know what you think about my comments

Added the recommended changes. Hopefully they cover the contexts you were concerned with.

Agree that for contexts where the value covers the entire structure, we can get the necessary context from the conversion function, e.g. in a backtrace / panic dump.

rmsyn:
> Let me know what you think of my review. It feels to me like the errors containing field names are not that necessary.

I can appreciate that reviews are incremental, but this could have been made clearer in the last set of changes.

The `field` entry is useful when viewing the error messages, e.g. if users have some form of logging enabled.

It also clarifies situations where multiple fields could return/bubble-up an error. For instance, in a function parsing the `permission` **AND** `range` field, how would you differentiate which field errors without attaching a debugger/`panic`ing?

romancardenas:
Hey @rmsyn sorry if it annoyed you. After the last changes, I thought that backtracking would be enough to find out which field changed, with no need to add static strings to the binary. In any case, if you think that variants that indicate the invalid field are necessary/valuable, we can definitely leave them as is, I have no strong feelings against them.

I will leave a couple of days in case anyone from @rust-embedded/riscv wants to leave their review and merge it to master.
Again, thank you so much for your contribution!

romancardenas (approved):

rmsyn:
> After the last changes, I thought that backtracking would be enough to find out which field changed, with no need to add static strings to the binary. In any case, if you think that variants that indicate the invalid field are necessary/valuable, we can definitely leave them as is, I have no strong feelings against them.

I think the `field` is helpful, but not exactly necessary. I can see the argument that including static strings in `riscv` could lead to an increase in static memory. An alternative that would take up less space would be an enum of all CSRs with fields. However, that seemed even more burdensome for maintenance, and downstream users.

If we did use an enum, or follow `svd` to use a module-specific error type, maybe we can use minimal formatting in `riscv`. Then users can decide how verbose they would like to get with error messages. The module-specific approach would require a reworking of the entire `result` module.

We could also just drop the `field` like you suggested, and have less directly helpful error messages. This may not be that large of a trade-off. My main concern is when working on a platform with only `printf`-style debugging available, e.g. I'm currently working on a platform without a working JTAG interface.

romancardenas:
Let's leave the `field` thing for now and open a new issue to discuss this in more depth. I'll add this PR to the merge queue tonight, so we can also advance in #211

PR #225, opened 2024-08-08
riscv: fix mtvec address field

rmsyn:
The `address` (BASE from the spec) needs to be 4-byte aligned, but the spec does not specify that the field is read/written as the masked address.

Shifts the value in `mtvec::write(addr, mode)`, and `Mtvec::address` for a more intuitive API.


rmsyn:
This fails in QEMU, wish the spec was more explicit about how the `BASE` field is encoded...

Issue #229, opened 2024-10-19
riscv: redefine CSR registers with new helper macros


rmsyn:
Existing CSR registers could be redefined using the new helper macros to gain extra functionality, and consistent definitions across the crate.

Not sure if this should go in before the next release, It's definitely not a blocking issue, and can be included in a future release.

romancardenas:
I would like to do a release soon with all the new changes so people can start playing with it and detect any potential issues. We can leave these changes for the next release (hopefully before 2025). In fact, what do you think about nominating issues and working on them for the next milestone?

rmsyn:
> In fact, what do you think about nominating issues and working on them for the next milestone?

That sounds great! Would you like me to create an issue for each CSR register, just to keep things modular?

Alternatively, I could add checkboxes to this issue with each CSR, and check them off as implementations are completed.

romancardenas:
I nominated a few issues to be part of the riscv 0.13.0 milestone.

In each issue, we should write a checkbox list to keep track of advances. We can also open a new issue (e.g., riscv 0.13.0), and each of us can assign one issue to work on.

Another option is using GitHub projects. It is a kind of Trello but strongly coupled to the repo.

Let me know what you prefer!

rmsyn:
> In each issue, we should write a checkbox list to keep track of advances.

This works for me. I'll go through the crate, and add a checkbox for each CSR.

romancardenas:
Currently, when compiling `riscv` for RV32 targets, I get the following warnings:

```
warning: unused variable: `uxl`
   --> riscv/src/register/mstatus.rs:418:31
    |
418 |     pub fn set_uxl(&mut self, uxl: XLEN) {
    |                               ^^^ help: if this is intentional, prefix it with an underscore: `_uxl`
    |
    = note: `#[warn(unused_variables)]` on by default

warning: unused variable: `sxl`
   --> riscv/src/register/mstatus.rs:443:31
    |
443 |     pub fn set_sxl(&mut self, sxl: XLEN) {
    |                               ^^^ help: if this is intentional, prefix it with an underscore: `_sxl`

warning: unused variable: `endianness`
   --> riscv/src/register/mstatus.rs:468:31
    |
468 |     pub fn set_sbe(&mut self, endianness: Endianness) {
    |                               ^^^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_endianness`

warning: unused variable: `endianness`
   --> riscv/src/register/mstatus.rs:492:31
[...]

PR #232, opened 2024-10-20
riscv: define marchid using CSR macros

rmsyn:
Adds a helper macro to test if a CSR is implemented on the platform.

Uses CSR helper macros to define the `marchid` register.

Related: #229

### DIFF
--- a/riscv/src/register/macros.rs
+++ b/riscv/src/register/macros.rs
@@ -92,6 +92,27 @@ macro_rules! read_csr_as {
             })
         }
     };
+
+    ($register:ident, $csr_number:literal, $sentinel:tt) => {
+        $crate::read_csr!($csr_number);
+
+        /// Reads the CSR.
+        ///
+        /// **WARNING**: panics on targets not implementing this CSR.
+        #[inline]
+        pub fn read() -> $register {
+            try_read().unwrap()
+        }
+
+        /// Attempts to reads the CSR.
+        #[inline]
+        pub fn try_read() -> $crate::result::Result<$register> {
+            match unsafe { _try_read()? } {
+                $sentinel => Err($crate::result::Error::Unimplemented),
+                bits => Ok($register { bits }),
+            }
+        }
+    };
 }
 
 /// `RV32`: Convenience macro to read a CSR register value as a `register` type.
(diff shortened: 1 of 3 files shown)


riscv/src/register/macros.rs
```diff
@@ -1028,3 +1028,20 @@ macro_rules! write_only_csr_field {
...
+/// Convenience macro for checking if a CSR matches a sentinel value.
+///
+/// If the CSR value matches the sentinel, the CSR is not implemented.
+#[macro_export]
+macro_rules! csr_is_implemented {
+    ($csr:ident, $sentinel:expr) => {
+        $crate::paste! {
+            impl $csr {
+                #[doc = " When [" $csr "] is hardwired to `" $sentinel "` it means that the CSR isn't implemented."]
+                pub const fn is_implemented(&self) -> bool {
```
romancardenas, line 1041: This makes me wonder: should we add a `NotImplemented` error variant?

riscv/src/register/marchid.rs
```diff
@@ -1,28 +1,9 @@
...
-    bits: NonZeroUsize,
-}
-
-impl Marchid {
-    /// Returns the contents of the register as raw bits
-    #[inline]
-    pub fn bits(&self) -> usize {
-        self.bits.get()
-    }
+read_only_csr! {
```
romancardenas, line 3: In case we want this new error variant, we can add a new macro/macro variant that, when trying to read/write, checks if the CSR is implemented

rmsyn: Right, I went with the simpler implementation to open the conversation. I think encoding the semantics into the type may be better, as well.

To start, I think I'll just add the `Unimplemented` error variant, and the corresponding macro branch to the read-only helpers. The only registers using the `NonZeroUsize`/`Unimplemented` are all read-only.

riscv/src/register/macros.rs
```diff
@@ -92,6 +92,27 @@ macro_rules! read_csr_as {
             })
         }
     };
+
+    ($register:ident, $csr_number:literal, $sentinel:tt) => {
+        $crate::read_csr!($csr_number);
+
+        /// Reads the CSR.
+        ///
+        /// **WARNING**: panics on non-`riscv` targets.
```
romancardenas, line 101:
```suggestion
        /// **WARNING**: panics on targets not implementing this CSR.
```

romancardenas (changes requested):
Minor change. It looks to me!

romancardenas (approved):

jessebraham (approved):
LGTM!

PR #233, opened 2024-10-21
riscv: define mcause using CSR macros

rmsyn:
Uses CSR helper macros to define the `mcause` register.

Related: #229

### DIFF
--- a/riscv/CHANGELOG.md
+++ b/riscv/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Use CSR helper macros to define `marchid` register
 - Re-use `try_*` functions in `mcountinhibit`
+- Use CSR helper macros to define `mcause` register
 
 ## [v0.12.1] - 2024-10-20
 
--- a/riscv/src/register/mcause.rs
+++ b/riscv/src/register/mcause.rs
@@ -2,32 +2,41 @@
 
 pub use crate::interrupt::Trap;
 
-/// mcause register
-#[derive(Clone, Copy, Debug)]
-pub struct Mcause {
-    bits: usize,
+read_only_csr! {
+    /// `mcause` register
+    Mcause: 0x342,
+    mask: 0xffff_ffff,
 }
 
-impl From<usize> for Mcause {
-    #[inline]
-    fn from(bits: usize) -> Self {
-        Self { bits }
-    }
+#[cfg(target_arch = "riscv32")]
+read_only_csr_field! {
+    Mcause,
+    /// Returns the `code` field.
+    code: [0:30],
+}
+
+#[cfg(not(target_arch = "riscv32"))]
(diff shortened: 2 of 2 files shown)


romancardenas (commented):
Can't we use a target-dependent `const INTERRUPT_BIT` set to 31 or 63 and forget about all these `#[cfg(target_arch=...)]`? I guess that to support that we would need to tweak the macros once again...

(This suggestion is not a strong one, if you feel that it is not necessary, I'm OK with the current approach)

rmsyn:
> Can't we use a target-dependent const INTERRUPT_BIT set to 31 or 63 and forget about all these #[cfg(target_arch=...)]?

So, the problem I ran into here is that the macro helpers currently use `literal`s for bitfield indexes (single, repeated-range, and field-range).

I attempted to use the `tt` fragment specifier, but it causes matching issues with the branch arms. It is too generic, and ends up matching with delimiting characters.

Using `tt` seems to work for the invocations in `mcountinhibit`, but fails with the test cases in `riscv/src/register/tests/*`.

The `expr` fragment specifier also doesn't work, because of the `$bit_start ..= $bit_end` and `[$bit_start :  $bit_end]` syntax.

Using `item` is too exclusive, because it leaves out `literal`.

That leaves us with using `literal`, and the somewhat awkward implementation syntax you pointed out.

If you know some other macro-fu to resolve the issue, I am happy to refactor.

> (This suggestion is not a strong one, if you feel that it is not necessary, I'm OK with the current approach)

I would also prefer more flexibility, but it seems this would require doubling the existing macro arms to include those that use `item`, and those that use `literal`. This would be for each arm in the `*csr_field` macros.

rmsyn:
> Can't we use a target-dependent const INTERRUPT_BIT set to 31 or 63 and forget about all these #[cfg(target_arch=...)]?

So, the problem I ran into here is that the macro helpers currently use `literal`s for bitfield indexes (single, repeated-range, and field-range).

I attempted to use the `tt` fragment specifier, but it causes matching issues with the branch arms. It is too generic, and ends up matching with delimiting characters.

Using `tt` seems to work for the invocations in `mcountinhibit`, but fails with the test cases in `riscv/src/register/tests/*`.

The `expr` fragment specifier also doesn't work, because of the `$bit_start ..= $bit_end` and `[$bit_start :  $bit_end]` syntax.

Using `item` is too exclusive, because it leaves out `literal`.

That leaves us with using `literal`, and the somewhat awkward implementation syntax you pointed out.

If you know some other macro-fu to resolve the issue, I am happy to refactor.

> (This suggestion is not a strong one, if you feel that it is not necessary, I'm OK with the current approach)

I would also prefer more flexibility, but it seems this would require doubling the existing macro arms to include those that use `tt`, and those that use `literal`. This would be for each arm in the `*csr_field` macros.

romancardenas (dismissed):
Let's leave it as is

romancardenas (approved):

PR #235, opened 2024-10-27
riscv: define medeleg using CSR macros

rmsyn:
Uses CSR helper macros to define the `medeleg` register.

Adds basic unit tests for the `medeleg` register.

Related: #229


rmsyn:
> I understand that previously there were no setters in the `Medeleg`, right?

Correct, there were no setters. I didn't know if this was intentional, because of the `set_` and `clear_` free functions.

`medeleg` is a read-write register from the spec, so I figured adding the setters was the way to go. Am I mistaken?

> BTW: I like the approach of adding simple unit tests for these register proxies

:heart:

romancardenas:
Probably there weren't getter/setters because of the atomic `set_*` functions. However, with these new setters via a proxy, we can modify the `medeleg` CSR with just one write to the actual register.

Issue #239, opened 2024-11-04
`riscv-rt`: reconfigurable trap frame


romancardenas:
We need to reconfigure the trap frame via feature flags. This is especially relevant for:

1. RV32E targets (see #189)
2. ESP32Cx targets (see [this issue in `esp-hal`](https://github.com/esp-rs/esp-hal/issues/2390))

Additionally, I foresee another breaking change: changing the order of the trap frame to follow the `x` order (from `x0` to `x31`). As far as I know, this is required for ESP32Cx devices. Maybe @bjoernQ can confirm this to us.

Issue #189, opened 2024-02-23
`riscv-rt`: Assembly algorithm for RAM init incompatible with upcoming RVE extension (future proofing)


hegza:
The RAM init function is incompatible with (as of yet, unratified) RV32E base ISA.

https://github.com/rust-embedded/riscv/blob/f5a2da9a05ebe710d8fc24b38939305de37ecf99/riscv-rt/src/asm.rs#L147-L158

The algorithm uses temporaries t0--t3 to copy data from from flash to RAM. However, the RV32E removes registers `x16`--`x31`, which includes `t3` / `x28`, therefore making this algorithm unavailable on RVE.

## Proposals for a solution

1. We could pick another register in range `x0--x16` in place of `t3` to make the code compatible between RVI and RVE
2. We could `#[cfg(riscve)]` the algorithm to use another register on RV32E.

romancardenas:
I personally prefer using another register that fits all the specifications

romancardenas:
Solved in #243

PR #243, opened 2024-11-09
`riscv-rt`: compatibility with RV32E and RV64E

romancardenas:
Avoids using `x16`+ registers in assembly code. This is necessary to provide compatibility with RV32E.
Note that there is still work to do. Namely, the trap frame must adapt to the target base instruction set.

Solves #189 

Related: #239

### DIFF
--- a/riscv-rt/src/lib.rs
+++ b/riscv-rt/src/lib.rs
@@ -547,21 +547,8 @@ use riscv::register::scause as xcause;
 #[cfg(not(feature = "s-mode"))]
 use riscv::register::mcause as xcause;
 
-pub use riscv_rt_macros::{entry, exception, external_interrupt, pre_init};
-
 pub use riscv_pac::*;
-
-#[cfg(all(target_arch = "riscv32", riscve))]
-pub use riscv_rt_macros::core_interrupt_rv32e as core_interrupt;
-#[cfg(all(target_arch = "riscv32", riscvi))]
-pub use riscv_rt_macros::core_interrupt_rv32i as core_interrupt;
-#[cfg(all(target_arch = "riscv64", riscve))]
-pub use riscv_rt_macros::core_interrupt_rv64e as core_interrupt;
-#[cfg(any(
-    all(target_arch = "riscv64", riscvi),
-    not(any(target_arch = "riscv32", target_arch = "riscv64"))
-))]
-pub use riscv_rt_macros::core_interrupt_rv64i as core_interrupt;
+pub use riscv_rt_macros::{core_interrupt, entry, exception, external_interrupt, pre_init};
 
 /// We export this static with an informative name so that if an application attempts to link
 /// two copies of riscv-rt together, linking will fail. We also declare a links key in
--- a/riscv-rt/src/lib.rs
+++ b/riscv-rt/src/lib.rs
@@ -551,12 +551,17 @@ pub use riscv_rt_macros::{entry, exception, external_interrupt, pre_init};
 
 pub use riscv_pac::*;
 
-#[cfg(target_arch = "riscv32")]
-pub use riscv_rt_macros::core_interrupt_riscv32 as core_interrupt;
-#[cfg(target_arch = "riscv64")]
-pub use riscv_rt_macros::core_interrupt_riscv64 as core_interrupt;
-#[cfg(not(any(target_arch = "riscv32", target_arch = "riscv64")))]
-pub use riscv_rt_macros::core_interrupt_riscv64 as core_interrupt; // just for docs, tests, etc.
+#[cfg(all(target_arch = "riscv32", riscve))]
+pub use riscv_rt_macros::core_interrupt_rv32e as core_interrupt;
+#[cfg(all(target_arch = "riscv32", riscvi))]
+pub use riscv_rt_macros::core_interrupt_rv32i as core_interrupt;
+#[cfg(all(target_arch = "riscv64", riscve))]
+pub use riscv_rt_macros::core_interrupt_rv64e as core_interrupt;
+#[cfg(any(
+    all(target_arch = "riscv64", riscvi),
+    not(any(target_arch = "riscv32", target_arch = "riscv64"))
+))]
+pub use riscv_rt_macros::core_interrupt_rv64i as core_interrupt;
 
 /// We export this static with an informative name so that if an application attempts to link
 /// two copies of riscv-rt together, linking will fail. We also declare a links key in
@@ -608,9 +613,9 @@ pub struct TrapFrame {
     #[cfg(not(riscve))]
     /// `x17`: argument register `a7`. Used to pass the eighth argument to a function.
     pub a7: usize,
-    #[cfg(riscve)]
+    #[cfg(all(target_arch = "riscv32", riscve))]
     _reserved0: usize,
-    #[cfg(riscve)]
+    #[cfg(all(target_arch = "riscv32", riscve))]
     _reserved1: usize,
 }
 
--- a/riscv-rt/src/lib.rs
+++ b/riscv-rt/src/lib.rs
@@ -532,7 +532,7 @@
 #![no_std]
 #![deny(missing_docs)]
 
-#[cfg(riscv)]
+#[cfg(any(target_arch = "riscv32", target_arch = "riscv64"))]
 mod asm;
 
 #[cfg(not(feature = "no-exceptions"))]
@@ -551,11 +551,11 @@ pub use riscv_rt_macros::{entry, exception, external_interrupt, pre_init};
 
 pub use riscv_pac::*;
 
-#[cfg(riscv32)]
+#[cfg(target_arch = "riscv32")]
 pub use riscv_rt_macros::core_interrupt_riscv32 as core_interrupt;
-#[cfg(riscv64)]
+#[cfg(target_arch = "riscv64")]
 pub use riscv_rt_macros::core_interrupt_riscv64 as core_interrupt;
-#[cfg(not(riscv))]
+#[cfg(not(any(target_arch = "riscv32", target_arch = "riscv64")))]
 pub use riscv_rt_macros::core_interrupt_riscv64 as core_interrupt; // just for docs, tests, etc.
 
 /// We export this static with an informative name so that if an application attempts to link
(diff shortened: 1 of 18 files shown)


riscv-rt/src/lib.rs
```diff
@@ -567,25 +567,40 @@ pub use riscv_rt_macros::core_interrupt_riscv64 as core_interrupt; // just for d
...
+    /// `x13`: argument register `a3`. Used to pass the fourth argument to a function.
     pub a3: usize,
+    /// `x14`: argument register `a4`. Used to pass the fifth argument to a function.
     pub a4: usize,
+    /// `x15`: argument register `a5`. Used to pass the sixth argument to a function.
     pub a5: usize,
+    /// `x16`: argument register `a6`. Used to pass the seventh argument to a function.
     pub a6: usize,
+    /// `x17`: argument register `a7`. Used to pass the eighth argument to a function.
     pub a7: usize,
```
rmsyn, line 604: Would it be a good idea to order these fields by there `x*` name (similar to the spec)?

Also, maybe we use coniditional compilation to not include the `x16`-`x31` registers for RV32E builds?

Mostly thinking about downstream users that may use a pointer to the `TrapFrame`, which they may or may not try to map to memory layout.

It may not be necessary, but could be a helpful reminder that those registers don't exist on RV32E platforms.

romancardenas: I'd try to avoid changing the order of the attributes in case users use C or assembly code that relies on the current order.

I'm working on feature-gating registers for RVE32. It's a bit tricky, as we need to add "trash" padding to ensure that the stack is 16-byte aligned. I'll open a PR shortly with all this.

rmsyn: > I'd try to avoid changing the order of the attributes in case users use C or assembly code that relies on the current order.

I guess that's also part of what I'm asking, "should we intentionally break downstream users compiling RV32E?" Especially regarding not providing them in-memory registers that don't exist on the platform.

For non-RV32E users, I agree with you, we should avoid breaking them without a good reason.

> I'm working on feature-gating registers for RVE32. It's a bit tricky, as we need to add "trash" padding to ensure that the stack is 16-byte aligned. I'll open a PR shortly with all this.

Sounds good, I'll keep an eye out for it.

Maybe for the missing registers on RV32E targets:

```rust
pub struct TrapFrame {
    #[cfg(not(rv32e))]
    pub a6: usize,
    #[cfg(rv32e)]
    _a6: usize,
    ...
}
```

rmsyn:
> My main challenge now is configuring riscv-rt-macros according to flags triggered in riscv-rt. I don't know how to do this without features

If this specifically about `RV32E` stuff, maybe a combination of `target_arch` and `target_feature` flags in a `#[cfg(...)]`? Although, it looks like there are [some restrictions on using `target_feature` on `riscv*` targets](https://doc.rust-lang.org/reference/attributes/codegen.html#riscv32-or-riscv64).

I would be happy to help out, if you have some work-in-progress code you would like extra eyes on.

romancardenas:
This is the best way I could think of. I tried forwarding environment variables from build scripts to no avail, as procedural macros are compiled separately from regular crates. The macro code can be squeezed a bit, but more or less this is it.

rmsyn:
> This is the best way I could think of. I tried forwarding environment variables from build scripts to no avail, as procedural macros are compiled separately from regular crates. The macro code can be squeezed a bit, but more or less this is it.

Yeah, I tried using conditional compilation with `target_feature = "i"` + `target_feature = "e"`, but it doesn't look like base ISA are available.

If `target_feature = "i"` and `target_feature = "e"` were available, we could use conditional compilation to define a `core_interrupt_inner` the four variants we care about, i.e. `riscv32e`, `riscv32i`, `riscv64e`, `riscv64i`. Then we could have the single `core_interrupt` macro function wrapping the unsafe inner functions.

I can look at the `rustc/cargo` source, and see what it takes to add base ISA as conditional compilation attribute. It may make more sense for the base ISA stuff to go under [the instruction_set attribute](https://doc.rust-lang.org/reference/attributes/codegen.html#the-instruction_set-attribute) (currently only has ARM in the docs).

The changes here LGTM.

rmsyn (approved):
> This is the best way I could think of. I tried forwarding environment variables from build scripts to no avail, as procedural macros are compiled separately from regular crates. The macro code can be squeezed a bit, but more or less this is it.

Yeah, I tried using conditional compilation with `target_feature = "i"` + `target_feature = "e"`, but it doesn't look like base ISA are available.

If `target_feature = "i"` and `target_feature = "e"` were available, we could use conditional compilation to define a `core_interrupt_inner` the four variants we care about, i.e. `riscv32e`, `riscv32i`, `riscv64e`, `riscv64i`. Then we could have the single `core_interrupt` macro function wrapping the unsafe inner functions.

I can look at the `rustc/cargo` source, and see what it takes to add base ISA as conditional compilation attribute. It may make more sense for the base ISA stuff to go under [the instruction_set attribute](https://doc.rust-lang.org/reference/attributes/codegen.html#the-instruction_set-attribute) (currently only has ARM in the docs).

The changes here LGTM.

romancardenas:
I tried something like that, but I found out that the procedural macros are compiled for the host, not the target. Thus, the `cfg(target_arch = "riscv32")` does not work unless your host machine is a RV32

romancardenas:
I'm currently with a considerable workload at work. Will come back to this this weekend, probably

rmsyn:
> I'm currently with a considerable workload at work. Will come back to this this weekend, probably

No worries, I hope all goes well.

romancardenas:
@rmsyn take a look to my latest commit. Using `cfg!(target =...)` didn't work, but I managed to do something very similar with environment variables.

riscv-rt/src/lib.rs
```diff
@@ -567,26 +559,51 @@ pub use riscv_rt_macros::core_interrupt_riscv64 as core_interrupt; // just for d
...
+    /// `x15`: argument register `a5`. Used to pass the sixth argument to a function.
     pub a5: usize,
+    #[cfg(not(riscve))]
+    /// `x16`: argument register `a6`. Used to pass the seventh argument to a function.
     pub a6: usize,
+    #[cfg(not(riscve))]
+    /// `x17`: argument register `a7`. Used to pass the eighth argument to a function.
     pub a7: usize,
+    #[cfg(all(target_arch = "riscv32", riscve))]
+    _reserved0: usize,
```
rmsyn, line 604: Are these reserved fields added for alignment?

riscv-rt/src/lib.rs
```diff
@@ -567,26 +559,51 @@ pub use riscv_rt_macros::core_interrupt_riscv64 as core_interrupt; // just for d
...
+    #[cfg(all(target_arch = "riscv32", riscve))]
+    _reserved1: usize,
```
rmsyn, line 606: Are these reserved fields added for alignment?

rmsyn (approved):
Changes LGTM, congrats on the `env` solution!

romancardenas:
I added a few minor improvements and wrote all the important changes in the CHANGELOG.md file

riscv-rt/src/lib.rs
romancardenas: Yes, the trap frame must be 16-byte aligned. Thus, for RV32E, we need a couple of `usize` reserved fields

rmsyn (approved):

romancardenas:
@hegza can you confirm that this version fits your RV32E projects?

riscv-rt/src/lib.rs
hegza: > I'm working on feature-gating registers for RVE32. It's a bit tricky, as we need to add "trash" padding to ensure that the stack is 16-byte aligned. I'll open a PR shortly with all this.

Wait, 16-byte aligned? You mean for RV32E? Stack on RV32E is specified to be 32-bit / 4-byte aligned. It's a specific feature of RV32E, ref: https://github.com/llvm/llvm-project/blob/528bcf3a55ca520c31c77ed5fbacf09bff8f39ec/clang/lib/Basic/Targets/RISCV.h#L145-L150.

hegza:
I maintain a short patch-set that works for us, here: https://github.com/rust-embedded/riscv/compare/master...hegza:riscv:feat/rve

I can smoke test our platform possibly today already.

romancardenas:
Thanks for your review! I fixed the alignment issue. Now, it should be basically identical to your fork.

hegza:
Tested RT by swapping it in place of our fork for a minimal smoke test and indeed there was no smoke. So, works on [my machine](https://github.com/soc-hub-fi/atalanta), I guess 😅  Compiles alright and prints over serial without putting the core into crazy states.

hegza (approved):

hegza:
For the record, if the RT manages to create any invalid stores  to registers `x16..` on an RV32E Rust backend, it would fail to compile. So, the fact that it compiles is already a good sign. TrapFrame looked alright as well.

romancardenas:
So I guess it is ready to merge, then!

romancardenas:
I just noticed that we were breaking the 16 byte alignment in our runtime. I guess we could add a bunch of feature gates to relax these constraints in RVxE targets, but I'd rather leave this work for a future PR

rmsyn (approved):

rmsyn:
> I guess we could add a bunch of feature gates to relax these constraints in RVxE targets, but I'd rather leave this work for a future PR

As far as I can tell, there is no difference for the stack pointer alignment in `RVxE` implementations, the only difference is the register count (based on Ch. 3 of the Unprivileged Spec). Is there somewhere else that specifies a difference in stack pointer alignment?

hegza:
I haven't found this specified either but that's what clang and LLVM are doing, as linked in earlier discussion already:

- https://github.com/llvm/llvm-project/blob/528bcf3a55ca520c31c77ed5fbacf09bff8f39ec/clang/lib/Basic/Targets/RISCV.h#L145-L150
- https://github.com/llvm/llvm-project/blob/5b17f85a7d722439a39f1ac1c554aed7858adab4/llvm/lib/Target/RISCV/RISCVTargetMachine.cpp#L145-L146
- https://github.com/llvm/llvm-project/blob/5b17f85a7d722439a39f1ac1c554aed7858adab4/llvm/lib/Target/RISCV/RISCVFrameLowering.cpp#L30-L36

We also discussed this earlier over at rust-lang/rust: https://github.com/rust-lang/rust/pull/130555#issuecomment-2362843912

rmsyn:
> I haven't found this specified either but that's what clang and LLVM are doing, as linked in earlier discussion already:
> 
>     * https://github.com/llvm/llvm-project/blob/528bcf3a55ca520c31c77ed5fbacf09bff8f39ec/clang/lib/Basic/Targets/RISCV.h#L145-L150
> 
>     * https://github.com/llvm/llvm-project/blob/5b17f85a7d722439a39f1ac1c554aed7858adab4/llvm/lib/Target/RISCV/RISCVTargetMachine.cpp#L145-L146
> 
>     * https://github.com/llvm/llvm-project/blob/5b17f85a7d722439a39f1ac1c554aed7858adab4/llvm/lib/Target/RISCV/RISCVFrameLowering.cpp#L30-L36
> 
> 
> We also discussed this earlier over at rust-lang/rust: [rust-lang/rust#130555 (comment)](https://github.com/rust-lang/rust/pull/130555#issuecomment-2362843912)

Looks like GCC does the same thing: https://github.com/gcc-mirror/gcc/blob/d136fa00f0d5faff8397edcd7e4ebb3445ab21b0/gcc/config/riscv/riscv.h#L202-L205

I wonder where they get this information from? Since they both do it, they must be basing it on something. I'm looking into some soft-core implementations like [CVW](https://github.com/openhwgroup/cvw) and [PicoRV32](https://github.com/YosysHQ/picorv32), since those look like the closest thing to an actual hardware implementation at the moment.

PR #244, opened 2024-11-13
riscv: define misa using CSR macros

rmsyn:
Removes the unused field range arguments from the CSR field enum helper macro.

Uses CSR helper macros to define the `misa` register.

Corrections for other users of the `XLEN` enum.

Adds basic unit tests for the `misa` register.

Related: #229


romancardenas (approved):
LGTM!

PR #245, opened 2024-12-03
riscv: define mstatus using CSR macro

rmsyn:
Adds an optional field documentation macro argument to allow adding documentation to CSR field enum variants.

Current macros for CSR field enum accessors allow for the MSB and LSB to hold the same value (for single-bit enum fields).

Clippy defaults to `deny(clippy::eq_op)`. The simplest solution is to allow the lint. A more robust solution is to special-case equal operands with new macro arm.

Uses CSR macro helpers to define the `mstatus` CSR register.

Adds a CSR test field macro arm for enum fields.

Adds more unit test cases for `Mstatus` field accessors.

### DIFF
--- a/riscv/src/register/mstatus.rs
+++ b/riscv/src/register/mstatus.rs
@@ -2,356 +2,200 @@
 
 pub use super::misa::XLEN;
 use crate::bits::{bf_extract, bf_insert};
+#[cfg(target_arch = "riscv32")]
+use crate::result::Error;
+use crate::result::Result;
 
-/// mstatus register
-#[derive(Clone, Copy, Debug)]
-pub struct Mstatus {
-    bits: usize,
+#[cfg(not(target_arch = "riscv32"))]
+read_write_csr! {
+    /// mstatus register
+    Mstatus: 0x300,
+    mask: 0x8000_0000_007f_fffe,
 }
 
-impl From<usize> for Mstatus {
-    #[inline]
-    fn from(bits: usize) -> Self {
-        Self { bits }
+#[cfg(target_arch = "riscv32")]
+read_write_csr! {
+    /// mstatus register
+    Mstatus: 0x300,
+    mask: 0x807f_fffe,
+}
+
+csr_field_enum! {
+    /// Additional extension state
+    XS {
+        default: AllOff,
+        /// All off
+        AllOff = 0,
+        /// None dirty or clean, some on
+        NoneDirtyOrClean = 1,
+        /// None dirty, some clean
+        NoneDirtySomeClean = 2,
+        /// Some dirty
+        SomeDirty = 3,
     }
 }
 
-/// Additional extension state
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
-pub enum XS {
-    /// All off
-    AllOff = 0,
-
-    /// None dirty or clean, some on
-    NoneDirtyOrClean = 1,
-
-    /// None dirty, some clean
-    NoneDirtySomeClean = 2,
-
-    /// Some dirty
-    SomeDirty = 3,
+csr_field_enum! {
+    /// Floating-point extension state
+    FS {
+        default: Off,
+        Off = 0,
+        Initial = 1,
+        Clean = 2,
+        Dirty = 3,
+    }
 }
 
-/// Floating-point extension state
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
-pub enum FS {
-    Off = 0,
-    Initial = 1,
-    Clean = 2,
-    Dirty = 3,
+csr_field_enum! {
+    /// Vector extension state
+    VS {
+        default: Off,
+        Off = 0,
+        Initial = 1,
+        Clean = 2,
+        Dirty = 3,
+    }
 }
 
-/// Vector extension state
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
-pub enum VS {
-    Off = 0,
-    Initial = 1,
-    Clean = 2,
-    Dirty = 3,
+csr_field_enum! {
+    /// Machine Previous Privilege Mode
+    MPP {
+        default: User,
+        User = 0,
+        Supervisor = 1,
+        Machine = 3,
+    }
 }
 
-/// Machine Previous Privilege Mode
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
-pub enum MPP {
-    Machine = 3,
-    Supervisor = 1,
-    User = 0,
+csr_field_enum! {
+    /// Supervisor Previous Privilege Mode
+    SPP {
+        default: User,
+        User = 0,
+        Supervisor = 1,
+    }
 }
 
-/// Supervisor Previous Privilege Mode
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
-pub enum SPP {
-    Supervisor = 1,
-    User = 0,
-}
-
-/// Non-instruction-fetch memory endianness
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
-pub enum Endianness {
-    BigEndian = 1,
-    LittleEndian = 0,
+csr_field_enum! {
+    /// Non-instruction-fetch memory endianness
+    Endianness {
+        default: LittleEndian,
+        LittleEndian = 0,
+        BigEndian = 1,
+    }
 }
 
 impl From<bool> for Endianness {
-    fn from(value: bool) -> Self {
-        match value {
-            true => Self::BigEndian,
+    fn from(val: bool) -> Self {
+        match val {
             false => Self::LittleEndian,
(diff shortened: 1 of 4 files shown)


riscv/src/register/mstatus.rs
```diff
@@ -643,11 +531,40 @@ mod test {
     #[test]
     fn test_mpp() {
```
rmsyn, line 532: Missed renaming the test-case. Should probably change to `test_mstatus` for clarity, and will update with any review changes requested.

riscv/src/register/mstatus.rs
```diff
@@ -361,64 +205,76 @@ impl Mstatus {
...
+    /// Whether either the FS field or XS field signals the presence of some dirty state
+    sd: 63,
+}
 
+impl Mstatus {
     /// Effective xlen in U-mode (i.e., `UXLEN`).
     ///
     /// In RISCV-32, UXL does not exist, and `UXLEN` is always [`XLEN::XLEN32`].
+    ///
+    /// **NOTE**: panics on RISCV-32 platforms.
```
romancardenas, line 241:
```suggestion
```

riscv/src/register/mstatus.rs
```diff
@@ -441,9 +297,24 @@ impl Mstatus {
     /// affect the mstatus CSR itself.
     #[inline]
```
romancardenas, line 298:
```suggestion
    ///
    /// **NOTE**: panics on RISCV-32 platforms.
    #[inline]
```

riscv/src/register/mstatus.rs
```diff
@@ -477,44 +366,43 @@ impl Mstatus {
...
-            () => Endianness::from(bf_extract(self.bits, 37, 1) != 0),
         }
     }
+
     /// Update M-mode non-instruction-fetch memory endianness
     ///
     /// Note this updates a previously read [`Mstatus`] value, but does not
     /// affect the mstatus CSR itself. See [`set_mbe`] to directly update the
     /// CSR.
     #[inline]
```
romancardenas, line 381:
```suggestion
    ///
    /// **NOTE**: panics on RISCV-32 platforms.
    #[inline]
```

romancardenas: Maybe in this kind of methods, for RV32, we should redirect users to use `mstatush`

romancardenas (changes requested):
minor changes. Thanks!

riscv/src/register/mstatus.rs
rmsyn: > we should redirect users to use `mstatush`

I agree, that would be helpful. I'll add it for `sbe` and `mbe`. These are the only two fields with overlap, right?

riscv/src/register/mstatus.rs
```diff
@@ -466,9 +338,30 @@ impl Mstatus {
...
+            () => {
+                self.bits = bf_insert(self.bits, 36, 1, endianness as usize);
+                Ok(())
+            }
+            #[cfg(target_arch = "riscv32")]
+            () => {
+                // SAFETY: `mstatush` is available on RISCV-32, and the value is safe to write.
+                // TODO: convert to `Mstatush` with `mbe` set
+                unsafe { super::mstatush::set_sbe(endianness) };
+                Ok(())
```
rmsyn, line 363: > we should redirect users to use `mstatush`

So, I just use the free functions in `mstatush` here, and for `try_set_mbe`. These are kind of a special-case.

I think when I make the changes for `mstatush`, I'll convert these functions to consume the `Mstatus` register, and return an `Mstatush` instance with the relevant field set.

It may be overengineering at that point, though. Especially since there are only these two fields. It may make more sense to just handle it internally here with:

```rust
let mut m = mstatush::read();
m.set_sbe(endianness);
mstatush::write(m);
Ok(())
```
I removed the panic note, because these `try_set` functions aren't fallible now. It may even make sense to remove the `try_` variants.

What do you think @romancardenas?

romancardenas: I am not sure about this because there is a difference in behavior between RV64 and RV32. In RV64, you modify a register copy that will later be applied to the real register. However, in RV32, you are directly modifying the register.

We can either:

1. Use the `read_composite_csr!` macro to deal with `mstatus` and `mstatush` together (check `mcycle`, for instance)
2. Just panic in `mstatus` and refer to `mstatush` in the docs

rmsyn: > Use the read_composite_csr! macro to deal with mstatus and mstatush together (check mcycle, for instance)

So, the `read_{m,s}be` is already handled, and there is no corresponding `write_composite_csr!`.

> However, in RV32, you are directly modifying the register.

I also had issue with this, which is why I suggested a function that consumes `Mstatus`, and returns an `Mstatus` and `Mstatush` instance. It would look something like:

```rust
pub fn set_mbe(self, endianness: Endianness) -> (Mstatus, Mstatush) {
    match () {
        #[cfg(not(target_arch = "riscv32"))]
        () => (
            Mstatus { bits: bf_insert(self.bits, 37, 1, endianness as usize) },
            Mstatush::new()
        ),
        () => (
            self,
            Mstatush { bits: bf_insert(0, 5, 1, endianness as usize },
        )
    }
}
```

Alternatively, to have a more idiomatic `set` function:

```rust
pub fn set_mbe(&mut self, endianness: Endianness) -> Mstatush {
    match () {
        #[cfg(not(target_arch = "riscv64"))]
        () => {
            self.bits = bf_insert(self.bits, 37, 1, endianness as usize);
            Mstatush::new()
        }
        () => Mstatush { bits: bf_insert(0, 5, 1, endianness as usize },
    }
}
```

It's a little cumbersome, but it avoids an unnecessary `panic`.

We could also explain the slightly awkward interface in the documentation.

> Just panic in mstatus and refer to mstatush in the docs

I would truly like to avoid as many runtime `panic`s as possible.

romancardenas: The thing is that `Mstatush` does not exist in RV64, so returning a non-existing register does not make sense for these targets. A regular setter works.

In RV32, you cannot set MBE nor SBE, as these fields do not belong to `Mstatus`, but to `Mstatush`. Thus, I am more inclined to return an `Unimplemented` error than provide a weird API that returns another register. I would rather write in the docs that in RV64 these methods are infallible but in RV32 they fail, as the MSTATUS register does not include those fields.

We can not include the setters for RISCV32 and only leave them for RV64. While the API will be slightly different depending on your target, well, this is in fact different depending on your target and we are making this difference explicit.

Note that, even if we remove these methods for RV32, we should add a brief explanation in the docs to make users aware of it. After all, `docs.rs` is built for a RV64 target.

rmsyn: > The thing is that Mstatush does not exist in RV64, so returning a non-existing register does not make sense for these targets. A regular setter works.

Right, which is why it returns the `Option`. On RISCV-64, it always returns `None`. Regardless, it is a pretty wonky `setter` API, so I'll change it for consistency.

> We can not include the setters for RISCV32 and only leave them for RV64. While the API will be slightly different depending on your target, well, this is in fact different depending on your target and we are making this difference explicit.

I like this approach, because other than the platform difference, there is no reason to have a `try` variant (it's an infallible API). Also, any users targetting RISCV-64 **and** RISCV-32 with the same code base will get an explicit "this function does not exist" error at compile-time (instead of the runtime error/panic). Maybe we could even provide a custom compile error to indicate to RISCV-32 users they should use `Mstatush` instead?

romancardenas: I like the idea of adding these methods to RV32 but only to throw a compile-time error redirecting users to `mstatush`!

rmsyn: > I like the idea of adding these methods to RV32 but only to throw a compile-time error redirecting users to mstatush!

So, I tried using a `compile_error` macro on the RV32 arm of the relevant functions, but that just causes the RV32 build to fail. Not sure how to produce a compile-error-on-use for a function. In other words, define the function for RV64 and RV32, but produce a compile-time error if RV32 users try to use the function.

The best I can think of is not defining the function for RV32 targets, and provide a note in the docs pointing users to `Mstatush` (which is what the current set of changes does). The other alternative is defining the function, but always returning the `Unimplemented` error.

romancardenas: I think the best approach is to not define these methods for RV32, as you did in your latest commit.

Maybe we can do the same with `set_uxl` and `set_sxl`? They are in a very similar situation. The only difference is that these methods are not part of another CSR, they just do not exist.

rmsyn: > Maybe we can do the same with set_uxl and set_sxl?

Alright, I added a commit with those changes. I left the getters, because it makes sense semantically to return the fixed `XLEN` values on RV32 (even if the fields technically don't exist).

PR #246, opened 2024-12-07
`riscv`: align assembly functions with `cortex-m`

### LINKED ISSUE #181: `riscv`: align assembly functions with `cortex-m` crate
Motivated by #179 (and #180), I took a look on how the `cortex-m` crate deals with [assembly instructions](https://github.com/rust-embedded/cortex-m/blob/master/cortex-m/src/asm.rs), and most of them are safe and use options to avoid the compiler from using the stack etc.

@rust-embedded/riscv what do you think about updating our assembly functions to follow the same fashion as in `cortex-m`? @rust-embedded/cortex-m any advice on this?

romancardenas:
Closes #181 

# On instructions that trigger exceptions

I decided to keep instructions that trigger exceptions (e.g., `ecall`) `unsafe` for their potential side effects. For example, if no exception handler is defined, triggering an exception can cause the program to crash. Also, raising exceptions is a potential way to avoid critical sections. Thus, one must be careful with them.

In any case, I'm open to to discuss this in case anyone has a different point of view.

### DIFF
--- a/riscv/src/asm.rs
+++ b/riscv/src/asm.rs
@@ -1,30 +1,24 @@
 //! Assembly instructions
 
 macro_rules! instruction {
-    ($(#[$attr:meta])*, unsafe $fnname:ident, $asm:expr) => (
+    ($(#[$attr:meta])*, unsafe $fnname:ident, $asm:expr, $($options:tt)*) => (
         $(#[$attr])*
-        #[inline]
+        #[inline(always)]
         pub unsafe fn $fnname() {
-            match () {
-                #[cfg(riscv)]
-                () => core::arch::asm!($asm),
-
-                #[cfg(not(riscv))]
-                () => unimplemented!(),
-            }
+            #[cfg(any(target_arch = "riscv32", target_arch = "riscv64"))]
+            core::arch::asm!($asm, $($options)*);
+            #[cfg(not(any(target_arch = "riscv32", target_arch = "riscv64")))]
+            unimplemented!();
         }
     );
-    ($(#[$attr:meta])*, $fnname:ident, $asm:expr) => (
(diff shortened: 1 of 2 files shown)


riscv/src/asm.rs
```diff
@@ -98,38 +110,18 @@ instruction!(
...
+    not(any(target_arch = "riscv32", target_arch = "riscv64")),
+    allow(unused_variables)
+)]
+pub fn sfence_vma(asid: usize, addr: usize) {
+    #[cfg(any(target_arch = "riscv32", target_arch = "riscv64"))]
+    unsafe {
+        core::arch::asm!("sfence.vma {0}, {1}", in(reg) addr, in(reg) asid, options(nostack));
+    };
+    #[cfg(not(any(target_arch = "riscv32", target_arch = "riscv64")))]
+    unimplemented!();
```
rmsyn, line 124: This can be expressed using the `instruction` macro.

I have an example of how to do it here: https://github.com/rust-embedded/riscv/compare/asm...rmsyn:riscv:asm

rmsyn:
> I decided to keep instructions that trigger exceptions (e.g., ecall) unsafe for their potential side effects.  For example, if no exception handler is defined, triggering an exception can cause the program to crash.

I think this is the right decision, too. Even if there is an exception handler defined, it could be defined using `unsafe` Rust, C, or assembly. Inherently `unsafe`, since we can't guarantee at compile-time what exception-triggering code calls into.

riscv/src/asm.rs
romancardenas: It didn't work on my machine, looks like it matches the first macro branch and assumes all the tokens correspond to the `options` stream.

rmsyn (approved):

PR #250, opened 2024-12-18
Simplify register macros

romancardenas:
I added a `cfg` field and some recursive macro calls to reduce duplicates in register macros.


rmsyn (approved):
Changes LGTM, this is a good cleanup!

PR #251, opened 2024-12-19
Make all CSR writes unsafe by default

### LINKED ISSUE #209: `riscv`: All the CSR write operations should be unsafe by default
In general we should assume that writing to CSRs could do something that potentially violates the Rust abstract model.

Macros like `write_csr_as` and `write_csr_as_usize` should at least default to `unsafe`, and maybe have an option to make a safe variant on a CSR by CSR basis.

romancardenas:
Closes #209 

**Work in Progress!** I added a `safe` pattern to opt out of unsafety in cases where we consider there to be no safety issues. I will list all the different registers where we use this macro and nominate some to maintain write safety.

### DIFF
--- a/riscv/CHANGELOG.md
+++ b/riscv/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- Make all CSR writes `unsafe` by default (#209)
 - Simplify register macros with `cfg` field
 - Align assembly functions with `cortex-m`
 - Use CSR helper macros to define `marchid` register
--- a/riscv/src/register/macros.rs
+++ b/riscv/src/register/macros.rs
@@ -12,13 +12,13 @@ macro_rules! read_csr {
         /// Reads the CSR.
         ///
         /// **WARNING**: panics on non-`riscv` targets.
-        #[inline]
+        #[inline(always)]
         unsafe fn _read() -> usize {
             _try_read().unwrap()
         }
 
         /// Attempts to read the CSR.
-        #[inline]
+        #[inline(always)]
         unsafe fn _try_read() -> $crate::result::Result<usize> {
             match () {
                 #[cfg($($cfg),*)]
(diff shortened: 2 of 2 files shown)


rmsyn:
Since most of these registers are `WARL`, what are we using for `safe` vs. `unsafe` criteria? Is it going to be related to side-effects, e.g. `mtvec` pointing to a potentially invalid memory region only triggered by an interrupt/exception?

Whatever criteria we come up with, maybe we should also add a blurb to the top-level `register` module about them?

Adding the `safe` opt-out is a good approach in my opinion. I agree that the CSR writes should be `unsafe` by default.

romancardenas:
> what are we using for `safe` vs. `unsafe` criteria? Is it going to be related to side-effects, e.g. `mtvec` pointing to a potentially invalid memory region only triggered by an interrupt/exception?

Yes, let us leave all writes of CSRs that might trigger invalid states unsafe. `mtvec` is a good example. I guess that `mstatus` must also be unsafe, as it can break critical sections, for instance.

romancardenas:
I have been reading the ISA and I feel like most of the registers should be `unsafe` to write. I think it would be a better idea force all write functions to be unsafe and open a new issue/RFC to nominate registers to be `safe`. What do you think @rust-embedded/riscv ?

rmsyn:
> I think it would be a better idea force all write functions to be unsafe and open a new issue/RFC to nominate registers to be safe.

I agree with this approach, and we should probably outline criteria for CSR write safety.

As we discussed before, most of the CSRs in the ISA are WARL (Write Anything Read Legal), so safety should be based on side-effects of writing to the registers.

For a rough set of attributes to consider:

- does a CSR write introduce potential memory safety issues in safe code?
- does a CSR write introduce potential undefined behavior in safe code?
- does a CSR write invalidate invariants in safe code?


It may be a good idea to add some sort of entry in either the top-level or `register` documentation. Something we can at least point contributors to when a `safe` RFC is being reviewed.

rmsyn (approved):
> I think it would be a better idea force all write functions to be unsafe and open a new issue/RFC to nominate registers to be safe.

I agree with this approach, and we should probably outline criteria for CSR write safety.

As we discussed before, most of the CSRs in the ISA are WARL (Write Anything Read Legal), so safety should be based on side-effects of writing to the registers.

For a rough set of attributes to consider:

- does a CSR write introduce potential memory safety issues in safe code?
- does a CSR write introduce potential undefined behavior in safe code?
- does a CSR write invalidate invariants in safe code?

It may be a good idea to add some sort of entry in either the top-level or `register` documentation. Something we can at least point contributors to when a `safe` RFC is being reviewed.

romancardenas:
I drafted a quick RFC template to ease the nomination process in #256

romancardenas:
Should we merge this PR?

rmsyn (approved):

PR #252, opened 2024-12-21
riscv: define mtvec CSR with macro helpers

rmsyn:
Uses the CSR macro helpers to define the `mtvec` CSR register.

Adds basic unit-tests for the `mtvec` CSR.

Ignores the warning for the `clippy::identity_op` to allow using ranged bitfield CSR macro arguments with a leading `0`.

Changes the macro argument type to `expr` to allow using constants, expressions, and other initializers for CSR mask values.

### DIFF
--- a/riscv/src/register/mtvec.rs
+++ b/riscv/src/register/mtvec.rs
@@ -1,50 +1,44 @@
 //! mtvec register
 
-/// mtvec register
-#[derive(Clone, Copy, Debug)]
-pub struct Mtvec {
-    bits: usize,
+const MASK: usize = usize::MAX;
+const TRAP_MASK: usize = 0b11;
+
+read_write_csr! {
+    /// mtvec register
+    Mtvec: 0x305,
+    mask: MASK,
 }
 
-/// Trap mode
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
-pub enum TrapMode {
-    Direct = 0,
-    Vectored = 1,
+csr_field_enum! {
+    /// Trap mode
+    TrapMode {
+        default: Direct,
+        Direct = 0,
+        Vectored = 1,
+    }
+}
+
+read_write_csr_field! {
+    Mtvec,
+    /// Accesses the trap-vector mode..
+    trap_mode,
+    TrapMode: [0:1],
 }
 
 impl Mtvec {
-    /// Returns the contents of the register as raw bits
-    #[inline]
-    pub fn bits(&self) -> usize {
-        self.bits
-    }
-
     /// Returns the trap-vector base-address
     #[inline]
-    pub fn address(&self) -> usize {
-        self.bits - (self.bits & 0b11)
+    pub const fn address(&self) -> usize {
+        self.bits - (self.bits & TRAP_MASK)
     }
 
-    /// Returns the trap-vector mode
+    /// Sets the trap-vector base-address.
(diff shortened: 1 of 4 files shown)


riscv/src/register/mtvec.rs
```diff
@@ -1,50 +1,64 @@
...
     #[inline]
-    pub fn trap_mode(&self) -> Option<TrapMode> {
-        let mode = self.bits & 0b11;
-        match mode {
-            0 => Some(TrapMode::Direct),
-            1 => Some(TrapMode::Vectored),
-            _ => None,
-        }
+    pub fn set_address(&mut self, address: usize) {
+        self.bits = (address & !TRAP_MASK) | (self.bits & TRAP_MASK);
```
rmsyn, line 42: This 4-byte alignment is in reference to the CSR specification, however in [`riscv-rt::asm`](https://github.com/rust-embedded/riscv/blob/master/riscv-rt/src/asm.rs#L245-L248) there is a reference to the trap table being 16-byte aligned.

Should we force 16-byte alignment here by masking the lower 4 bits?

riscv/src/register/mtvec.rs
```diff
@@ -1,50 +1,64 @@
...
+    TrapMode {
+        default: Direct,
+        Direct = 0,
+        Vectored = 1,
     }
+}
 
+read_write_csr_field! {
+    Mtvec,
+    /// Accesses the trap-vector mode..
```
romancardenas, line 23:
```suggestion
    /// Accesses the trap-vector mode.
```

riscv/src/register/mtvec.rs
```diff
@@ -1,50 +1,64 @@
...
+    TrapMode: [0:1],
+}
+
+impl Mtvec {
     /// Returns the trap-vector base-address
     #[inline]
-    pub fn address(&self) -> usize {
-        self.bits - (self.bits & 0b11)
+    pub const fn address(&self) -> usize {
+        self.bits - (self.bits & TRAP_MASK)
```
romancardenas, line 32: This is just a matter of tastes, but in case you also prefer this approach:
```suggestion
        self.bits & !TRAP_MASK
```

riscv/src/register/mtvec.rs
romancardenas: According to the RISCV Priv spec:

> The value in the BASE field must always be aligned on a 4-byte boundary, and the MODE setting may impose additional alignment constraints on the value in the BASE field.
...
An implementation may have different alignment constraints for different modes. In particular, MODE=Vectored may have stricter alignment constraints than MODE=Direct. Allowing coarser alignments in Vectored mode enables vectoring to be implemented without a hardware adder circuit.

So, without additional information about the target, we can only ensure that BASE is 4-byte aligned. For further versions of riscv-rt, we can think of a mechanism to apply a custom alignment depending on the target. For example, the E310x chip expects 64-byte alignment when working in vectored mode.

romancardenas: In this way, we can add a `try_set_address` that checks that `base` is at least 4-byte aligned

romancardenas (changes requested):

romancardenas (dismissed):

romancardenas:
Looks like Clippy added new lints. @rmsyn could you address them?

PR #255, opened 2025-01-11
riscv: define mvendorid CSR with macro helpers

rmsyn:
Uses CSR macro helpers to define the `mvendorid` CSR register.

### DIFF
--- a/riscv/src/register/mvendorid.rs
+++ b/riscv/src/register/mvendorid.rs
@@ -22,3 +22,37 @@ read_only_csr_field! {
     /// The encoded value returned by `offset` does not include the odd parity bit (`0x80`).
     offset: [0:6],
 }
+
+impl Mvendorid {
+    /// Represents the JEDEC manufacture continuation byte.
+    pub const CONTINUATION: u8 = 0x7f;
+
+    /// Gets the decoded JEDEC manufacturer ID from the `mvendorid` value.
+    ///
+    /// # Note
+    ///
+    /// This function returns an iterator over the decoded bytes.
+    ///
+    /// An iterator is needed because the encoding can theoretically return a max count (`0x1ff_ffff`) of continuation bytes (`0x7f`).
+    ///
+    /// The final byte in the iterator is the `offset`, including the odd parity bit (set only if even).
+    pub fn jedec_manufacturer(&self) -> impl Iterator<Item = u8> {
+        const DONE: usize = usize::MAX;
+
+        let mut bank = self.bank();
+        let offset = self.offset();
+
+        core::iter::from_fn(move || match bank {
+            DONE => None,
+            0 => {
+                bank = DONE;
+                let parity = ((1 - (offset.count_ones() % 2)) << 7) as usize;
+                Some((parity | offset) as u8)
+            }
+            _ => {
+                bank -= 1;
+                Some(Self::CONTINUATION)
+            }
+        })
+    }
+}
--- a/riscv/src/register/mvendorid.rs
+++ b/riscv/src/register/mvendorid.rs
@@ -1,34 +1,24 @@
 //! mvendorid register
 
-use core::num::NonZeroUsize;
-
-/// mvendorid register
-#[derive(Clone, Copy, Debug)]
-pub struct Mvendorid {
-    bits: NonZeroUsize,
+read_only_csr! {
+    /// `mvendorid` register
+    Mvendorid: 0xF11,
+    mask: 0xffff_ffff,
+    sentinel: 0,
 }
 
-impl Mvendorid {
-    /// Returns the contents of the register as raw bits
-    #[inline]
-    pub fn bits(&self) -> usize {
-        self.bits.get()
-    }
-
-    /// Returns the JEDEC manufacturer ID
-    #[inline]
-    pub fn jedec_manufacturer(&self) -> usize {
-        self.bits() >> 7
-    }
+read_only_csr_field! {
+    Mvendorid,
+    /// Represents the number of continuation bytes (`0x7f`) in the JEDEC manufacturer ID.
+    bank: [7:31],
 }
 
-read_csr!(0xF11);
-
-/// Reads the CSR
-#[inline]
-pub fn read() -> Option<Mvendorid> {
-    let r = unsafe { _read() };
-    // When mvendorid is hardwired to zero it means that the mvendorid
-    // csr isn't implemented.
-    NonZeroUsize::new(r).map(|bits| Mvendorid { bits })
+read_only_csr_field! {
+    Mvendorid,
+    /// Represents the final offset field in the JEDEC manufacturer ID.
+    ///
+    /// # Note
+    ///
+    /// The encoded value returned by `offset` does not include the odd parity bit (`0x80`).
+    offset: [0:6],
 }
--- a/riscv/src/register/mvendorid.rs
+++ b/riscv/src/register/mvendorid.rs
@@ -56,3 +56,45 @@ impl Mvendorid {
         })
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_mvendorid() {
+        (0..u32::BITS)
+            .map(|r| ((1u64 << r) - 1) as usize)
+            .for_each(|raw| {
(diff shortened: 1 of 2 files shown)


riscv/src/register/mvendorid.rs
```diff
@@ -1,34 +1,8 @@
...
-impl Mvendorid {
-    /// Returns the contents of the register as raw bits
-    #[inline]
-    pub fn bits(&self) -> usize {
-        self.bits.get()
-    }
-
-    /// Returns the JEDEC manufacturer ID
-    #[inline]
-    pub fn jedec_manufacturer(&self) -> usize {
```
romancardenas, line 20: Should we add the `jedec_manufacturer` field?

rmsyn: Good question, I missed that function when converting to the CSR macro.

After reading the ISA again, it looks like the `jedec_manufacturer` implementation was incorrect.


```
mvendorid encodes the number of one-byte continuation codes in the Bank field, and
encodes the final byte in the Offset field, discarding the parity bit. For example, the JEDEC
manufacturer ID 0x7f 0x7f 0x7f 0x7f 0x7f 0x7f 0x7f 0x7f 0x7f 0x7f 0x7f 0x7f 0x8a (twelve
continuation codes followed by 0x8a) would be encoded in the mvendorid CSR as 0x60a.
```

So, we would probably want to add `bank` and `offset` fields, along with a correct `jedec_manufacturer` function that decodes the ID:

```rust
pub const fn jedec_manufacturer(&self) -> (usize, usize) {
    (self.bank(), 0x80 | self.offset())
}
```

riscv/src/register/mvendorid.rs
```diff
@@ -1,34 +1,30 @@
...
-    // When mvendorid is hardwired to zero it means that the mvendorid
-    // csr isn't implemented.
-    NonZeroUsize::new(r).map(|bits| Mvendorid { bits })
+impl Mvendorid {
+    /// Gets the JEDEC manufacturer ID decoded into parts:
+    ///
+    /// - `bank`: the number of continuation bytes (`0x7f`)
+    /// - `offset`: specific offset in the JEDEC bank number
+    pub fn jedec_manufacturer(&self) -> (usize, usize) {
+        (self.bank(), (0x80 | self.offset()))
```
romancardenas, line 28: Instead of `0x80`, I think there goes the corresponding odd parity bit of `offset`. We can either include this logic or ignore the parity bit, as done in the `mvendorid` register. Another option is completely removing `jedec_manufacturer` and leaving `bank` and `offset`

rmsyn: > Instead of 0x80, I think there goes the corresponding odd parity bit of offset.

I'm not sure I understand what you mean here. The `0x80` **is** the odd parity bit added back to the `offset` number.

> Another option is completely removing `jedec_manufacturer` and leaving `bank` and `offset`

This is my preferred route. I think we should leave it up to users on how they want to encode the JEDEC manufacturer number.

I'll add some documentation to the `offset` field about the parity bit.

---

Not sure in practice, but the theoretical max number of continuation bytes is `0x1ff_ffff`. We obviously can't output such a number (`[0x7f; 0x1ff_ffff]`) in an embedded environment (not saying you suggested that, just saying). However, maybe some user wants to return an `Iterator` of continuation bytes :shrug:

I'm not so familiar with the use of JEDEC manufacturer numbers, and how they're usually handled in industry.

romancardenas: > The `0x80` **is** the odd parity bit added back to the `offset` number.

The odd parity bit is `0x80` only when the number of `1`s is even (e.g., `0x0a` becomes `0x8a`). But it should be `0x00` when the number of `1`s is already odd (e.g., `0x0b` remains `0x0b`).

>  However, maybe some user wants to return an `Iterator` of continuation bytes

It might be useful for someone to implement an iterator that shows the full JEDEC manufacturer number. If you are in the mood to implement it, then go ahead. Otherwise, we could act lazily and wait for someone who asks for this feature.

romancardenas (dismissed):
We can leave it as is or add an iterator for the JEDEC manufacturer number. Whatever you prefer @rmsyn

riscv/src/register/mvendorid.rs
rmsyn: > It might be useful for someone to implement an iterator that shows the full JEDEC manufacturer number. If you are in the mood to implement it, then go ahead. Otherwise, we could act lazily and wait for someone who asks for this feature.

I agree. I've implemented the `Iterator` solution, however there is a major `Git operations` outage on GitHub right now. I'll push my changes as soon as I can.

rmsyn:
> We can leave it as is or add an iterator for the JEDEC manufacturer number. Whatever you prefer @rmsyn

Please wait to merge the changes, I have an `Iterator` implementation for `jedec_manufacturer`, I'm just waiting on GitHub to resolve their current outage for `Git operations`.

riscv/src/register/mvendorid.rs
rmsyn: > The odd parity bit is 0x80 only when the number of 1s is even (e.g., 0x0a becomes 0x8a). But it should be 0x00 when the number of 1s is already odd (e.g., 0x0b remains 0x0b).

So, I've implemented the parity bit on the assumption that it was based on `offset` being even/odd. However, based on your comments, it should be based on if the `1`s count in `offset` is even or odd, is that right?

For example, an odd `offset` with an even amount of `1`s should have the parity bit set (`0x03` -> `0x83`)?

riscv/src/register/mvendorid.rs
```diff
@@ -1,34 +1,99 @@
...
+    /// This function returns an iterator over the decoded bytes.
+    ///
+    /// An iterator is needed because the encoding can theoretically return a max count (`0x1ff_ffff`) of continuation bytes (`0x7f`).
+    ///
+    /// The final byte in the iterator is the `offset`, including the odd parity bit (set only if even).
+    pub fn jedec_manufacturer(&self) -> impl Iterator<Item = u8> {
+        let mut done = false;
+        let mut bank = self.bank();
+        let offset = self.offset();
+        let parity = ((1 - (offset.count_ones() % 2)) << 7) as usize;
```
romancardenas, line 43:
```suggestion
```

riscv/src/register/mvendorid.rs
```diff
@@ -1,34 +1,99 @@
...
 
-    /// Returns the JEDEC manufacturer ID
-    #[inline]
-    pub fn jedec_manufacturer(&self) -> usize {
-        self.bits() >> 7
+        core::iter::from_fn(move || match bank {
+            0 if done => None,
+            0 => {
+                done = true;
+                Some((parity | offset) as u8)
```
romancardenas, line 49:
```suggestion
                let parity = ((1 - (offset.count_ones() % 2)) << 7) as usize;
                Some((parity | offset) as u8)
```

romancardenas (approved):
LGTM!

I suggested a small optimization that leaves computing the parity bit when needed. The idea is to reduce the size of the resulting iterator, as it captures one less value.

riscv/src/register/mvendorid.rs
romancardenas, line 43:
```suggestion
        let mut bank = self.bank() + 2; // add 2 for Some(offset) + None
        let offset = self.offset();
```

riscv/src/register/mvendorid.rs
```diff
@@ -1,34 +1,99 @@
...
+        let parity = ((1 - (offset.count_ones() % 2)) << 7) as usize;
 
-    /// Returns the JEDEC manufacturer ID
-    #[inline]
-    pub fn jedec_manufacturer(&self) -> usize {
-        self.bits() >> 7
+        core::iter::from_fn(move || match bank {
+            0 if done => None,
+            0 => {
+                done = true;
```
romancardenas, line 48:
```suggestion
            0 => None,
            1 => {
                bank -= 1;
                let parity = ((1 - (offset.count_ones() % 2)) << 7) as usize;
```

romancardenas (approved):
I suggested some niche optimizations to reduce the size of the resulting iterator

romancardenas (approved):

PR #257, opened 2025-01-19
riscv: define `pmpcfgx` types with CSR macros

rmsyn:
Uses CSR helper macros to define `pmpcfgx` field types.


romancardenas (approved):
LGTM

I say in the ISA that now there are `pmpcfg` registers from 1 to 15. Should we add them in a future PR?

rmsyn:
> Should we add them in a future PR?

I think this is a good idea. That way the discussions can focus entirely on their addition. Also, I think it will tie in with the ongoing work to make all CSR writes `unsafe`.

romancardenas:
@rmsyn please fix conflicts

romancardenas (approved):

PR #259, opened 2025-01-20
riscv-rt: Add RISCV_MTVEC_ALIGN to control the alignment of the vector table

ia0:
The RISC-V Privileged Specification (version 1.11) for mtvec CSR says:
> The value in the BASE field must always be aligned on a 4-byte boundary, and the MODE setting may
> impose additional alignment constraints on the value in the BASE field.
> [...]
> An implementation may have different alignment constraints for different modes. In particular,
> MODE=Vectored may have stricter alignment constraints than MODE=Direct.

The Ibex Reference Guide for mtvec CSR says[^1]:
> The trap-vector base address, always aligned to 256 bytes

Other implementations have yet other alignment constraints (see https://github.com/rust-embedded/riscv/pull/259#issuecomment-2602810245).

This PR introduces the `RISCV_MTVEC_ALIGN` environment variable to let users (in particular PACs) choose the alignment constraint appropriate to their implementation.

[^1]: https://ibex-core.readthedocs.io/en/latest/03_reference/cs_registers.html#machine-trap-vector-base-address-mtvec

### DIFF
--- a/riscv-rt/src/interrupts.rs
+++ b/riscv-rt/src/interrupts.rs
@@ -7,97 +7,24 @@
 //!
 //! In vectored mode (i.e., `v-trap` feature enabled), interrupt dispatching is handled by hardware.
 //! To support this mode, we provide inline assembly code that defines the interrupt vector table.
+//! Since the alignment constraint of this vector table is implementation-specific, it can be
+//! changed by setting the `RISCV_MTVEC_ALIGN` environment variable (the default is 4).
 //!
 //! # Note
 //!
-//! If your target has custom core interrupt sources, the target PAC might provide equivalent
-//! code to adapt for the target needs. In this case, you may need to opt out this module.
-//! To do so, activate the `no-interrupts` feature of the `riscv-rt` crate.
-
-#[cfg(not(feature = "v-trap"))]
-extern "C" {
-    fn SupervisorSoft();
-    fn MachineSoft();
-    fn SupervisorTimer();
-    fn MachineTimer();
-    fn SupervisorExternal();
-    fn MachineExternal();
-}
-
-/// Array with all the core interrupt handlers sorted according to their interrupt source code.
-///
-/// # Note
-///
-/// This array is necessary only in direct mode (i.e., `v-trap` feature disabled).
-#[cfg(not(feature = "v-trap"))]
-#[no_mangle]
-pub static __CORE_INTERRUPTS: [Option<unsafe extern "C" fn()>; 12] = [
-    None,
-    Some(SupervisorSoft),
-    None,
-    Some(MachineSoft),
-    None,
-    Some(SupervisorTimer),
-    None,
-    Some(MachineTimer),
-    None,
-    Some(SupervisorExternal),
-    None,
-    Some(MachineExternal),
-];
-
-/// It calls the corresponding interrupt handler depending on the interrupt source code.
-///
-/// # Note
-///
-/// This function is only required in direct mode (i.e., `v-trap` feature disabled).
-/// In vectored mode, interrupt handler dispatching is performed directly by hardware.
-///
-/// # Safety
-///
-/// This function must be called only from the [`crate::start_trap_rust`] function.
-/// Do **NOT** call this function directly.
-#[cfg(not(feature = "v-trap"))]
-#[inline]
-#[no_mangle]
-pub unsafe extern "C" fn _dispatch_core_interrupt(code: usize) {
-    extern "C" {
-        fn DefaultHandler();
-    }
-    match __CORE_INTERRUPTS.get(code) {
-        Some(Some(handler)) => handler(),
-        _ => DefaultHandler(),
-    }
-}
+//! If your target has custom core interrupt sources, the target PAC might provide equivalent code
+//! to adapt for the target needs (and is responsible for any alignment constraint). In this case,
+//! you may need to opt out this module. To do so, activate the `no-interrupts` feature of the
+//! `riscv-rt` crate.
 
 // In vectored mode, we also must provide a vector table
-#[cfg(all(
-    any(target_arch = "riscv32", target_arch = "riscv64"),
-    feature = "v-trap"
-))]
-core::arch::global_asm!(
-    r#" .section .trap, "ax"
-        .weak _vector_table
-        .type _vector_table, @function
-        
-        .option push
-        .balign 0x4 // TODO check if this is the correct alignment
-        .option norelax
-        .option norvc
-        
-        _vector_table:
-            j _start_trap                     // Interrupt 0 is used for exceptions
-            j _start_SupervisorSoft_trap
-            j _start_DefaultHandler_trap      // Interrupt 2 is reserved
-            j _start_MachineSoft_trap
-            j _start_DefaultHandler_trap      // Interrupt 4 is reserved
-            j _start_SupervisorTimer_trap
-            j _start_DefaultHandler_trap      // Interrupt 6 is reserved
-            j _start_MachineTimer_trap
-            j _start_DefaultHandler_trap      // Interrupt 8 is reserved
-            j _start_SupervisorExternal_trap
-            j _start_DefaultHandler_trap      // Interrupt 10 is reserved
-            j _start_MachineExternal_trap
-        
-        .option pop"#
-);
+#[riscv::pac_enum(unsafe CoreInterruptNumber)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+enum Interrupt {
+    SupervisorSoft = 1,
+    MachineSoft = 3,
+    SupervisorTimer = 5,
+    MachineTimer = 7,
+    SupervisorExternal = 9,
+    MachineExternal = 11,
+}
--- a/riscv/macros/src/lib.rs
+++ b/riscv/macros/src/lib.rs
@@ -265,7 +265,20 @@ impl PacEnumItem {
     }
 
     fn vector_table(&self) -> TokenStream2 {
-        let mut asm = String::from(
+        let align = match std::env::var("RISCV_MTVEC_ALIGN") {
+            Ok(x) => x.parse::<u32>().ok(),
+            Err(std::env::VarError::NotPresent) => Some(4),
+            Err(std::env::VarError::NotUnicode(_)) => None,
+        };
+        let align = match align {
+            Some(x) if x.is_power_of_two() && 4 <= x => x,
+            _ => {
+                return quote!(compile_error!(
+                    "RISCV_MTVEC_ALIGN is not a power of 2 (minimum 4)"
+                ))
+            }
+        };
+        let mut asm = format!(
             r#"
 #[cfg(all(feature = "v-trap", any(target_arch = "riscv32", target_arch = "riscv64")))]
 core::arch::global_asm!("
@@ -274,7 +287,7 @@ core::arch::global_asm!("
     .type _vector_table, @function
     
     .option push
-    .balign 0x4 // TODO check if this is the correct alignment
+    .balign {align}
     .option norelax
     .option norvc
     
@@ -315,6 +328,8 @@ core::arch::global_asm!("
         let max_discriminant = self.max_number;
         let valid_matches = self.valid_matches();
 
+        let is_core_interrupt = matches!(attr, PacTrait::Interrupt(InterruptType::Core));
+
         // Push the trait implementation
         res.push(quote! {
             unsafe impl riscv::#trait_name for #name {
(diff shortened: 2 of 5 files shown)


hegza:
Truly, each interrupt controller has its own alignment requirements. For instance, the [riscv-fast-interrupt specification (CLIC)](https://github.com/riscv/riscv-fast-interrupt) specifies 4-byte alignment in CLINT mode and 64-byte alignment in CLIC mode.

This has been an issue for us at https://github.com/soc-hub-fi/atalanta working on an Ibex-based CPU design as well, as not being able to parameterize this forces us to flag out & write the trap vector ourselves. I'd be in favor of a flag such as `mtvec-align-256`, though I don't know if that scales nicely.

romancardenas:
I think the best approach is using a linker symbol, such as _heap_size, for instance. By default, the symbol should be 4. PACs can overwrite it according to their particular needs.

The linker script should check that the alignment is a power of 2, I guess (and greater than 4)

rmsyn:
> I'm not sure how we can parametrize mtvec alignment without cargo features, so this PR simply uses the stricter alignment.

One possibility is to use const generics to parameterize the `Mtvec::set_address` function directly, or offer another `Mtvec::set_address_aligned<const N: usize>`.

Something like:

```rust
impl Mtvec {
    pub fn set_address_aligned<const N: usize>(&mut self, address: usize) {
        // check alignment does not violate spec requirements
        const _: () = assert!(N >= 4);
        const _: () = assert_eq!(N % 4, 0);
        const MASK: usize = usize::MAX & !((1usize << N) - 1);

        self.set_address(address & MASK);
    }
}
```

The above could also obviously be made into a fallible `try_set_address_aligned`, just used this one for brevity.

riscv-rt/src/interrupts.rs
```diff
@@ -81,7 +81,7 @@ core::arch::global_asm!(
         .type _vector_table, @function
         
         .option push
-        .balign 0x4 // TODO check if this is the correct alignment
+        .balign 0x100 // RISC-V requires 4, but implementations may be stricter
```
rmsyn, line 84: I really do not think we should add such strict alignment requirements based on a single 32-bit RISC-V core implementation.

rmsyn (changes requested):
> I'm not sure how we can parametrize mtvec alignment without cargo features, so this PR simply uses the stricter alignment.

One possibility is to use const generics to parameterize the `Mtvec::set_address` function directly, or offer another `Mtvec::set_address_aligned<const N: usize>`.

Something like:

```rust
impl Mtvec {
    pub fn set_address_aligned<const N: usize>(&mut self, address: usize) {
        // check alignment does not violate spec requirements
        const _: () = assert!(N >= 4);
        const _: () = assert_eq!(N % 4, 0);
        const MASK: usize = usize::MAX & !((1usize << N) - 1);

        self.set_address(address & MASK);
    }
}
```

The above could also obviously be made into a fallible `try_set_address_aligned`, just used this one for brevity.

---

For the actual alignment of the trap vector function, I think @romancardenas approach is the right way to go.

ia0:
> I'd be in favor of a flag such as `mtvec-align-256`, though I don't know if that scales nicely.

It kind of scales thanks to the `cfg_global_asm!` macro (which I had to move to crate-level). However, this has an impact on compile time since this expands the macro multiple times. But that's similar to what's done in `riscv-rt/src/asm.rs` anyway.

I used the values 4 (default), 16, 64, and 256, since 4, 64, and 256 are used, and 16 is the missing power of 2 with an even exponent.

> I think the best approach is using a linker symbol, such as _heap_size, for instance.

I don't think this works? `.balign` takes an absolute expression. Feel free to demonstrate what you had in mind with a PR if I misunderstood it.

> One possibility is to use const generics to parameterize the `Mtvec::set_address` function directly

This seems unrelated. This PR is about letting users configure the alignment of the vector table in `riscv-rt`. What you mention seems to be a functionality of `riscv` and already too late, since the problem is compile-time, not runtime.

riscv-rt/src/interrupts.rs
ia0: Added cargo features for control.

riscv-rt/src/lib.rs
```diff
@@ -532,6 +532,26 @@
 #![no_std]
 #![deny(missing_docs)]
 
+/// Parse cfg attributes inside a global_asm call.
+#[allow(unused_macros)]
+macro_rules! cfg_global_asm {
+    {@inner, [$($x:tt)*], } => {
+        core::arch::global_asm!{$($x)*}
```
ia0, line 539: Note: I moved this from `src/asm.rs` and did 2 changes:
- Fully qualify `core::arch::global_asm!()`
- All `unused_macros` since it's only used on `riscv32`, `riscv64`, or without `no-interrupts`.

romancardenas:
Perhaps another option is using a procedural macro with environment variables. Something like this: https://github.com/rust-embedded/riscv/blob/b391a4860ddf707324b26ceb9166aa52bc9556a9/riscv-rt/build.rs#L44

PACs would set the mtvec alignment as part of their `build.rs` script. Then, the `riscv-rt-macros` would generate the assembly code for the table according to that table. If the environment variable is not defined, the default alignment is 4.
The macro can check that the alignment is a power of 2, or any check that we consider necessary.

(I'm not against the approach of this PR, but I'd rather have something more configurable that would not potentially lead to more features etc.)

romancardenas:
THe `riscv-macros` crate already provides a similar macro:
https://github.com/rust-embedded/riscv/blob/b391a4860ddf707324b26ceb9166aa52bc9556a9/riscv/macros/src/lib.rs#L277

We can make a new version of this macro that also allows to define the alignment of the vector table and use it in `riscv-rt`,

romancardenas:
The good thing about the `riscv-macros` macro is that it serves for custom core interrupt enumeration, not only the enumeration provided by the RISC-V standard

ia0:
> THe `riscv-macros` crate already provides a similar macro:

Ah, I've seen this too late. I'll update the PR to reuse it.

ia0:
> > THe `riscv-macros` crate already provides a similar macro:
> 
> Ah, I've seen this too late. I'll update the PR to reuse it.

Actually, I don't think we can reuse it without creating a separate library for proc-macro helpers (or possibly making `riscv-rt` depend on `riscv-macros`). What do you think?

romancardenas:
Good (and delicate) question. Creating a new `riscv-util-macro` is an option, but not my favorite.

I personally don't want `riscv` to depend on `riscv-rt-macros`. Maybe someone wants to use `riscv` with a different runtime, and keeping them as much isolated as possible is desirable.. However, I don't see any issue in `riscv-rt-macros` depending on `riscv-macros`. After all, users already had already chosen to use `riscv-rt`.

Any thoughts/suggestions/concerns?

ia0:
> However, I don't see any issue in `riscv-rt-macros` depending on `riscv-macros`.

A proc-macro cannot depend on another proc-macro as a library. So it would only be `riscv-rt` depending on `riscv-macros`.

I updated the PR.

riscv-rt/Cargo.toml
```diff
@@ -24,6 +24,7 @@ riscv-target-parser = { path = "../riscv-target-parser", version = "0.1.0" }
 
 [dependencies]
 riscv = { path = "../riscv", version = "0.12.0" }
+riscv-macros = { path = "../riscv/macros", version = "0.1.0" }
```
romancardenas, line 27:
```suggestion
riscv = { path = "../riscv", version = "0.12.0", features = ["riscv-macros"] }
```

riscv-rt/src/interrupts.rs
```diff
@@ -75,29 +75,4 @@ pub unsafe extern "C" fn _dispatch_core_interrupt(code: usize) {
...
-            j _start_DefaultHandler_trap      // Interrupt 6 is reserved
-            j _start_MachineTimer_trap
-            j _start_DefaultHandler_trap      // Interrupt 8 is reserved
-            j _start_SupervisorExternal_trap
-            j _start_DefaultHandler_trap      // Interrupt 10 is reserved
-            j _start_MachineExternal_trap
-        
-        .option pop"#
-);
+riscv_macros::vector_table!();
```
romancardenas, line 78:
```suggestion
riscv::vector_table!();
```

In fact, `riscv-rt/src/interrupts.rs` can just simply use the `#[riscv::pac_enum(unsafe CoreInterruptNumber)]` macro over an enumeration. In that case, the `vector_table` macro is not necessary.

riscv-rt/Cargo.toml
ia0: That's actually a default feature, so I just removed the dependency.

riscv-rt/src/interrupts.rs
ia0: I was thinking about this, but there's the `weak`/`global` symbol issue. We would need to control this from an attribute.

ia0: That said, I can't find if we guarantee that `_vector_table` is weak in `riscv-rt`, so we might be able to make it strong instead. And then we can use `pac_enum`.

romancardenas: In this case, I'd leave it as a strong symbol. PACs should ensure that the table fits the target specifications and can enable the `riscv-rt/no-interrupts` feature to avoid `riscv-rt` from generating the table when needed.

romancardenas:
`__CORE_INTERRUPTS` and `_dispatch_core_interrupt` are also generated by the pac_enum macro:

https://github.com/rust-embedded/riscv/blob/cfb7128f7684cf35b1346b88d96d36325f8b8103/riscv/macros/src/lib.rs#L380

https://github.com/rust-embedded/riscv/blob/cfb7128f7684cf35b1346b88d96d36325f8b8103/riscv/macros/src/lib.rs#L386

They are only necessary when the `v-trap` feature is disabled, though the `pac_enum` macro does not include that feature gate currently.

I think the way to go is:

- Modifying the `pac_enum` macro so `__CORE_INTERRUPTS` and `_dispatch_core_interrupt` are only active when `v-trap` is disabled.
- Removing  `__CORE_INTERRUPTS` and `_dispatch_core_interrupt` from `riscv_rt::interrupts.rs` and leave the macro implementation.
- Do not feature-gate the `pac_enum` macro invocation in `riscv_rt::interrupts.rs`

romancardenas:
Actually, the first action point is not that immediate. `v-trap` feature gates only apply for `CoreInterruptNumber`. For other `ExceptionHandler` and `ExternalInterruptHandler`, there should not be feature gates.

riscv/macros/src/lib.rs
```diff
@@ -265,7 +265,25 @@ impl PacEnumItem {
...
+        let mut align = match std::env::var("RISCV_MTVEC_ALIGN") {
+            Ok(x) => x.parse::<u32>().ok(),
+            Err(std::env::VarError::NotPresent) => Some(4),
+            Err(std::env::VarError::NotUnicode(_)) => None,
+        };
+        if let Some(value) = align {
+            if !value.is_power_of_two() || value < 4 {
+                align = None;
+            }
+        }
```
rmsyn, line 277:
```suggestion
        let align = match std::env::var("RISCV_MTVEC_ALIGN") {
            Ok(x) => x.parse::<u32>().ok().and_then(|value| if !value.is_power_of_two() || value < 4 {
                None
            } else {
                Some(value)
            }),
            Err(std::env::VarError::NotPresent) => Some(4),
            Err(std::env::VarError::NotUnicode(_)) => None,
        };
```

rmsyn (commented):
Minor nit, removes the need for a mutable `align` variable.

riscv/macros/src/lib.rs
ia0: I did this on the second match for readability.

ia0:
> I think the way to go is:

I updated the PR accordingly (also gating the `extern` declarations since they were in `src/interrupts.rs`).

riscv-rt/src/interrupts.rs
```diff
@@ -14,64 +14,7 @@
 //! code to adapt for the target needs. In this case, you may need to opt out this module.
```
romancardenas, line 14: Could you add a few lines of documentation for vectored mode about the `RISCV_MTVEC_ALIGN` environment variable and why it is necessary? You should also mention that, in most of the cases, this environment variable must be handled by the PAC crate, not users directly

romancardenas (dismissed):
LGTM! I added a minor documentation suggestion.

I think we should add a directive to`riscv-rt`'s build script to re-compile the crate if `RISCV_MTVEC_ALIGN` changes its value, right after this:

https://github.com/rust-embedded/riscv/blob/b391a4860ddf707324b26ceb9166aa52bc9556a9/riscv-rt/build.rs#L49

We can opt-out this check if the `v-trap` vector is disabled (no vector table at all) or if the `no-interrupts` feature is enabled (i.e., a PAC is implementing its custom core interrupt vector).

PACs with custom core interrupt vectors (i.e., those that use `riscv::pac_enum(CoreInterruptNumber)` must also include a similar directive to their `build.rs`. We can also include this in the docs.

riscv-rt/src/interrupts.rs
ia0: I'm not good at documentation, so I attempted something. Feel free to suggest an alternative.

romancardenas (approved):
LGTM! Thanks for this PR, it adds lots of value to the ecosystem

romancardenas:
@ia0 could you squash commits before merging?

rmsyn (approved):

PR #270, opened 2025-02-19
riscv: make mcycle & minstret writable

hegza:
The RISC-V Instruction Set Manual Volume II: Privileged Architecture 20240411 specifies mcycle and minstret as "MRW" for read-write, instead of "MRO" for read-only. Therefore these registers should be writable.

Quick question: should there be also a `write_composite_csr` function for these registers since they are 64-bit registers even on 32-bit targets? That macro didn't exist so far but I could add it. Notably, just being able to write the lower bits is useful.

### DIFF
--- a/riscv/src/register/macros.rs
+++ b/riscv/src/register/macros.rs
@@ -449,6 +449,30 @@ macro_rules! read_composite_csr {
     };
 }
 
+/// Convenience macro to write a composite value to a CSR register.
+///
+/// - `RV32`: writes 32-bits into `hi` and 32-bits into `lo` to create a 64-bit value
+/// - `RV64`: writes a 64-bit value into `lo`
+#[macro_export]
+macro_rules! write_composite_csr {
+    ($hi:expr, $lo:expr) => {
+        /// Writes the CSR as a 64-bit value
+        #[inline]
+        pub unsafe fn write64(bits: u64) {
+            match () {
+                #[cfg(target_arch = "riscv32")]
+                () => {
+                    $hi((bits >> 32) as usize);
+                    $lo(bits as usize);
+                }
+
+                #[cfg(not(target_arch = "riscv32"))]
+                () => $lo(bits as usize),
+            }
+        }
+    };
+}
+
 macro_rules! set_pmp {
     () => {
         /// Set the pmp configuration corresponding to the index.
--- a/riscv/src/register/mcycle.rs
+++ b/riscv/src/register/mcycle.rs
@@ -3,4 +3,4 @@
 read_csr_as_usize!(0xB00);
 write_csr_as_usize!(0xB00);
 read_composite_csr!(super::mcycleh::read(), read());
-write_composite_csr!(|bits| super::mcycleh::write(bits), |bits| write(bits));
+write_composite_csr!(super::mcycleh::write, write);
--- a/riscv/src/register/mcycle.rs
+++ b/riscv/src/register/mcycle.rs
@@ -1,4 +1,5 @@
 //! mcycle register
 
 read_csr_as_usize!(0xB00);
+write_csr_as_usize!(0xB00);
 read_composite_csr!(super::mcycleh::read(), read());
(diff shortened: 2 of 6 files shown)


rmsyn (dismissed):
LGTM. After re-reading the privileged spec, these registers are WARL, so I'm guessing `XLEN=32` platforms would just mask the upper bits in hardware on writes.

> should there be also a write_composite_csr function for these registers since they are 64-bit registers even on 32-bit targets? That macro didn't exist so far but I could add it.

If you'd like to, I'd say go for it. Adding the redundancy to mask in software as well as hardware could be helpful.

Edit: although, since this macro is `write_csr_as_usize`, we would only be writing the lower 32-bits anyway, right?

riscv/src/register/macros.rs
```diff
@@ -449,6 +449,34 @@ macro_rules! read_composite_csr {
...
+        pub unsafe fn write64(bits: u64) {
+            match () {
+                #[cfg(target_arch = "riscv32")]
+                () => {
+                    // Setting low value to max first prevents the register from being temporarily
+                    // reduced by a transaction that is intended for increasing the total
+                    // value.
+                    $lo(usize::MAX);
+                    $hi((bits >> 32) as usize);
+                    $lo(bits as usize);
```
hegza, line 470: So, there's a little bit of a trick here that I took from a RISC-V specification related to `mtimer`. For `mtimer` this trick is very necessary to avoid mtimer driven compare matches that cause interrupts (e.g., any CLINT / base spec interrupt architecture).

I don't think mcycle and minstret natively drive any compare matches in a sensible implementation but it may or may not be a good idea to do it like this just be safe. What do you think?

Another valid--possibly better--interpretation would be to say that this macro just sets the registers and if you want to have specific guarantees you write the code manually.

hegza (commented):
Other than the above question, I think the macro & CSR implementation is done.

riscv/src/register/macros.rs
hegza: Technically, I guess it's possible you could want to have the opposite guarantee of "register should not temporarily increase for a transaction that is intended for decreasing the total value". I'm therefore predisposed against doing it like shown and instead just doing the writes naively.

rmsyn: I don't think implicitly writing a value to a CSR is a good idea. In this specific case, it may be safe, however, it could have different semantics depending on the CSR.

As an example, `mstatus` uses `1` values to enable interrupts. So, a user of `write64` would temporarily enable interrupts, even if explicitly writing values to disable interrupts. If an interrupt happens to fire in the short time between the two `$lo` writes, the system could be put into an inoperable state.

If `write_composite_csr` is intended to be general-use, I would suggest removing the implicit first write.

hegza:
Write macro now writes only the necessary writes as per advice. I'm inclined to merge.

rmsyn (approved):
LGTM!

riscv/src/register/mcycle.rs
```diff
@@ -1,4 +1,6 @@
 //! mcycle register
 
 read_csr_as_usize!(0xB00);
+write_csr_as_usize!(0xB00);
 read_composite_csr!(super::mcycleh::read(), read());
+write_composite_csr!(|bits| super::mcycleh::write(bits), |bits| write(bits));
```
romancardenas, line 6: Do you have any reason for not doing something like this?

```suggestion
write_composite_csr!(super::mcycleh::write, write);
```

riscv/src/register/minstret.rs
```diff
@@ -1,4 +1,6 @@
 //! minstret register
 
 read_csr_as_usize!(0xB02);
+write_csr_as_usize!(0xB02);
 read_composite_csr!(super::minstreth::read(), read());
+write_composite_csr!(|bits| super::minstreth::write(bits), |bits| write(bits));
```
romancardenas, line 6: Same comment here:

```suggestion
write_composite_csr!(super::minstreth::write, write);
```

romancardenas (dismissed):

romancardenas (approved):
LGTM!

hegza:
Someone with the correct permissions will need to push the merge button 🙂

PR #271, opened 2025-02-27
`riscv-rt`: add interrupts and exceptions to `link.x` only when required

romancardenas:
As now core interrupts and exceptions can be re-defined in PACs to adapt to different targets, I thought it would be a good idea to clean the linker script and only add weak bindings to standard exceptions and interrupts when these are not disabled via the `no-exceptions` and `no-interrupts` features.

Now, in the `build.rs` script, we check if the `no-interrupts` and `no-exceptions` features are enabled. If not, we copy-paste the content of the new `interrupts.x` and `exceptions.x` files, respectively.

This is not a breaking change, it should work the same way. In fact, it might help find bugs in PACs that define partially their custom interrupt or exceptions.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@nightly
+      - uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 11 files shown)


rmsyn (approved):
LGTM, I'll test on hardware this week.

romancardenas:
@rmsyn can you re-aprove this PR?

rmsyn (approved):

PR #273, opened 2025-03-04
Discard DefaultHandler weak symbol with LTO

### LINKED ISSUE #247: `riscv-rt`: Cannot be built with `lto = true` if the user defines replacement for `.weak` symbols
With the current design choice for some functions to be written in Assembly and be marked as `.weak`, this can remove the ability for downstream users to build with `lto = true`. I initially filed a bug with `rustc`, however, bjorn3 there suggested that this crate could replace `.weak` with `#[linkage = "weak"]` behind some sort of nightly feature. I have a locally patched version, however it at minimum would require `#![feature(linkage)]` and possibly even `#![feature(linkage, naked_functions)]`. I don't know where this project stands on nightly usage, but I think this could fairly easily be hidden behind a feature flag.

ia0:
Fixes #247

### DIFF
--- a/riscv-rt/src/asm.rs
+++ b/riscv-rt/src/asm.rs
@@ -36,6 +36,7 @@ riscv_rt_macros::llvm_arch_patch!();
 // and the global pointer. Then it calls _start_rust.
 cfg_global_asm!(
     ".section .init, \"ax\"
+    .lto_discard DefaultHandler
     .global _start
 
 _start:",


rmsyn (changes requested):
This breaks a number of builds.

Is there a reason you opened this over the fixes in https://github.com/rust-embedded/riscv/pull/254?

ia0:
> This breaks a number of builds.

Yes, I'm quite surprised. For my project it works both with and without LTO...

> Is there a reason you opened this over the fixes in #254?

No, actually #254 also fixes my problem. Closing this in favor of #254.

ia0:
It might be compiler version related. So it's not reliable, sad.

Issue #238, opened 2024-11-01
`riscv-rt`: adapt linker file to look more like `cortex-m-rt`'s


romancardenas:
I suggest adopting part of `cortex-m-rt` structure:

# Make `link.x` include `memory.x`

In this way, PACs and BSPs would have to provide a `memory.x` linker file that defines the memory layout and regions.

End users would need to do something like `RUSTFLAGS="-C link-arg=-Tlink.x` when building binaries, regardless of the final target.

# Keep using region aliases

I think the current `riscv-rt` approach is quite versatile and I don't see any advantage in limiting sections to FLASH and RAM.

# Use `__` to define start and end of sections

I think it is a good idea to follow the same naming convention as in `cortex-m-rt` (`__stext`, `__etext`...).

romancardenas:
More information in #236

PR #274, opened 2025-03-08
`riscv-rt`: include device in link.x

romancardenas:
This follows the approach of `cortex-m-rt`.

Related to #238


romancardenas:
Will try to work on it. It is a bit painful, as we need to make files visible to the linker. Maybe a `tests-build` crate? We cannot use the `tests` crate as `trybuild` and cross-compiled projects don't get along very well.

rmsyn:
> Will try to work on it. It is a bit painful, as we need to make files visible to the linker. Maybe a tests-riscv crate? We cannot use the tests crate as trybuild and cross-compiled projects don't get along very well.

No worries, I wasn't aware it was such a pain. I think the automated tests would definitely be very useful, but their lack is not necessarily a blocker for this set of changes.

Whatever you want to do.

romancardenas:
Ok! I will merge this, open a similar PR, and then work on a better test scenario so we can experiment with the new features.

PR #275, opened 2025-03-11
Add memory feature

romancardenas:
Very similar to #274 , but now the `memory` feature enables including `memory.x` directly in `link.x`.

Usually, a standard embedded rust project has three linker files:

- `link.x`: the foundational linker file of the runtime. In the RISC-V ecosystem, this linker file is provided by `riscv-rt`. It defines all the sections, checks that alignments are correct, etc.
- `device.x`: It is usually provided by PACs. It defines the device-specific interrupt sources, exceptions, etc.
- `memory.x`: It is usually provided by BSPs. It defines available memory regions on a device (e.g., FLASH and RAM).

In `cortex-m-rt`, the `memory.x` file is mandatory and always included in their linker file. Alternatively, PACs may enable the `device` feature to include `device.x` at the end of their `link.x` file. In this way, when compiling, users only need to provide the `-C link-arg=Tlink.x`. This applies to every `cortex-m` device, which makes this approach user-friendly and convenient.

With this PR, in `riscv-rt` it will be possible to have such behavior by enabling both the `device` and `memory` features. New PAC version would now enable the `device` feature, and new BSP version would now enable the `memory` feature. Users would only need to modify the `RUSTFLAGS` when compiling, as now they would only need to do  `-C link-arg=Tlink.x` for linking.

As mentioned in the PR, leaving these two features optionals have two major benefits:

1. Backwards compatibility.
2. We still allow unusual/experimental link strategies that do not rely on either a `device.x` or `memory.x` file without too much effort. This makes `riscv-rt` quite versatile.

Once this PR is merged, I would say that #238 is solved.


romancardenas:
Will work on a more versatile test framework right after this PR

rmsyn (approved):

PR #279, opened 2025-04-23
`_default_start_trap` and `_default_setup_interrupts`

romancardenas:
Still working on #247 

If I'm not wrong, after this PR is accepted, we only need to think on `_pre_init` and `_mp_hook`. I have a copuled of suggestions for these symbols, so I will leave them for further PRs.


rmsyn (approved):
LGTM.

PR #282, opened 2025-04-29
`riscv-rt`: Leave `__pre_init` under a feature gate

romancardenas:
Still working on #247 

The idea is to remove `__pre_init` from the default `riscv-rt` crate. If users want it, they must enable the `pre-init` feature.
If enabled, no default `__pre_init` symbol is provided, as we assume that users will develop their routine.

Additionally, the `riscv_rt::pre_init` attribute macro is now marked as deprecated, as running Rust code before RAM initialization is unsound, and assembly code is preferred.

### DIFF
--- a/riscv-rt/macros/src/lib.rs
+++ b/riscv-rt/macros/src/lib.rs
@@ -174,8 +174,15 @@ fn is_correct_type(ty: &Type, name: &str) -> bool {
 }
 
 /// Attribute to mark which function will be called at the beginning of the reset handler.
+/// You must enable the `pre_init` feature in the `riscv-rt` crate to use this macro.
 ///
-/// **IMPORTANT**: This attribute can appear at most *once* in the dependency graph. Also, if you
+/// # IMPORTANT
+///
+/// This attribute is **deprecated**, as it is not safe to run Rust code **before** the static
+/// variables are initialized. The recommended way to run code before the static variables
+/// are initialized is to use the `global_asm!` macro to define the `__pre_init` function.
+///
+/// This attribute can appear at most *once* in the dependency graph. Also, if you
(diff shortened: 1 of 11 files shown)


riscv-rt/macros/src/lib.rs
```diff
@@ -197,6 +204,7 @@ fn is_correct_type(ty: &Type, name: &str) -> bool {
 ///
 /// # fn main() {}
 /// ```
+#[deprecated(note = "Use global_asm! to define the __pre_init function instead")]
```
rmsyn, line 207: Will we be removing this in the next (or another future) point release?

I wasn't aware that using Rust code before RAM init was unsound. Is this true even for code that only uses the stack?

rmsyn (approved):

riscv-rt/macros/src/lib.rs
romancardenas: I'm not sure about how Rust's ABI can break something. I guess that, if you do not use static variables, it should be fine

romancardenas: From `cortex-m` documentation:

> A `#[pre_init]` macro is also provided to run a function before RAM initialisation, but its use is deprecated as it is not defined behaviour to execute Rust code before initialisation. It is still possible to create a custom `pre_init` function using assembly.

rmsyn: We also discussed this on Monday's meeting. I think the consensus was that there were too many preconditions + invariants to justify keeping a `pre_init` implementation in Rust.

I think it may still be helpful to point to the feature-gated `__pre_init` function in the `riscv-rt` docs as an example for users that would want to implement their own. Maybe in the place currently held by `#[pre_init]` or above the `_pre_init_trap`?

PR #283, opened 2025-04-30
Simplify `riscv::interrupt::machine::nested`

romancardenas:
Just simplify the `nested` function for M-mode, following the same fashion as for S-mode.


rmsyn (approved):
Right, with the `set_` functions using the RISCV asm to only update the relevant bitfield, it makes sense to just use those methods.

That was the original intention of updating the in-memory struct, so this simplification is the right idea.

PR #284, opened 2025-05-15
register: add `scontext` register 

rmsyn:
Adds the definition for the `scontext` Supervisor debug context CSR.

Adds basic unit tests for the `scontext` CSR.

Related: #1


romancardenas (approved):
LGTM

PR #289, opened 2025-05-17
`riscv-rt`: Remove weak symbol for _mp_hook

### LINKED ISSUE #247: `riscv-rt`: Cannot be built with `lto = true` if the user defines replacement for `.weak` symbols

romancardenas:
Closes #247


rmsyn (approved):
LGTM

PR #293, opened 2025-06-05
`riscv-peripheral`: utility function to get `MTIME` frequency from `MTIMER` peripheral

romancardenas:
While working on adapting the `e310x-hal` to this new approach, I found it would be useful to expose the `MTIME_FREQ` with a method to avoid using weird `riscv_peripheral::MTIMER::<Clint>::MTIME_FREQ` things.


rmsyn (approved):
LGTM.

Issue #295, opened 2025-06-07
Set _hart_stack_size = SIZEOF(.stack) for single-hart cores.


janderholm:
I got the following error while building for a riscv core.

```
  = note: some arguments are omitted. use `--verbose` to show all linker arguments
  = note: rust-lld: error:
          ERROR(riscv-rt): .stack section is too small for allocating stacks for all the harts.
          Consider changing `_max_hart_id` or `_hart_stack_size`.

          rust-lld: error:
          ERROR(riscv-rt): .stack section is too small for allocating stacks for all the harts.
          Consider changing `_max_hart_id` or `_hart_stack_size`.
```

I can set _hart_stack_size to a smaller value, but I wonder if it would make more sense to set it to the size of the stack if the "single-hart" feature is enabled? They should effectively be the same anyway shouldn't they? This seem to work in my local memory.x:

`_hart_stack_size = SIZEOF(.stack);`

janderholm:
Should the ASSERT be changed to `>=` as well? Or am I off?

```ASSERT(SIZEOF(.stack) > (_max_hart_id + 1) * _hart_stack_size, "
ERROR(riscv-rt): .stack section is too small for allocating stacks for all the harts.
Consider changing `_max_hart_id` or `_hart_stack_size`.");
```

Currently if I set _hart_stack_size to 2048 and have a stack of 2048 this will result in a linking error.

romancardenas:
Hi @janderholm ! Thanks for opening the issue!

I think your proposals are reasonable. Would you like to start a PR? Otherwise, I can try to implement your idea.

Note that `_hart_stack_size = SIZEOF(.stack);` must only be active when the `single-hart` feature is on. To do feature-dependent changes in the `link.x`, we use the `build.rs` script with `${FEATURE}` wildcards.

janderholm:
PR posted!

romancardenas:
Closed as per #296

PR #296, opened 2025-06-11
Correct hart_stack_size assertion

janderholm:
Solves #295.

Corrected the assert such that the stack can be divided exactly in the number of harts. Also set the default hart stack size to this value.

### DIFF
--- a/riscv-rt/CHANGELOG.md
+++ b/riscv-rt/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Main function no longer needs to be close to _start. A linker script may copy
   all code to RAM and keep .init in flash/ROM.
+- By default, the stack is now split into equal parts based on the number of
+  harts.
 
 ### Fixed
 
--- a/riscv-rt/link.x.in
+++ b/riscv-rt/link.x.in
@@ -72,7 +72,7 @@ ${INCLUDE_LINKER_FILES}
 PROVIDE(_stext = ORIGIN(REGION_TEXT));
 PROVIDE(_stack_start = ORIGIN(REGION_STACK) + LENGTH(REGION_STACK));
 PROVIDE(_max_hart_id = 0);
-PROVIDE(_hart_stack_size = 2K);
+PROVIDE(_hart_stack_size = SIZEOF(.stack) / (_max_hart_id + 1));
 PROVIDE(_heap_size = 0);
 
 SECTIONS
--- a/riscv-rt/src/lib.rs
+++ b/riscv-rt/src/lib.rs
@@ -90,7 +90,10 @@
 //!
 //! This symbol defines stack area size for *one* hart.
 //!
-//! If omitted this symbol value will default to 2K.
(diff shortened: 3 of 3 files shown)


riscv-rt/build.rs
```diff
@@ -11,6 +11,15 @@ fn add_linker_script(arch_width: u32) -> io::Result<()> {
     let mut content = fs::read_to_string("link.x.in")?;
     content = content.replace("${ARCH_WIDTH}", &arch_width.to_string());
 
+    // Single core devices can use the entire stack, no need to limit it.
+    let hart_stack_size = if env::var_os("CARGO_FEATURE_SINGLE_HART").is_some() {
+        "SIZEOF(.stack)"
+    } else {
+        "2K"
```
romancardenas, line 18: Would something like this make sense?

```suggestion
        "SIZEOF(.stack) / (_max_hart_id + 1)"
```

romancardenas (commented):
LGTM. Please, add the change in the `CHANGELOG.md`

riscv-rt/build.rs
rmsyn: Agreed, the environment variable check can probably be skipped.

rmsyn (approved):

riscv-rt/build.rs
janderholm: I think I agree too. I have a question though since I don't have much experience with multi core devices.

If we default to splitting the stack equal parts among the cores (4 in case of 4 cores), do we need to add some more checks?

I see there's a heap variable with a default size of 0. If someone have currently set the heap to something that will cause the stack of their last core to become smaller than the others.

Is this something that needs to be taken into account?

janderholm: Ahh, yes i see, because .heap does ` . += _heap_size;`

With heap set to 1024:
.heap                      0  0x1000008
.stack                  4088  0x1000008

With heap set to 1024:
.heap                   1024  0x1000008
.stack                  3064  0x1000408

I checked with `nm <elf> | grep _hart_stack_size` and it looks like it works even with sizes not perfectly divisable. But I'm unsure about the alignment. For example with 3 harts I get a stack size of 1362.

It looks like these instructions in _start might be there to fix that that though?

```
cfg_global_asm!(
    "la t1, _stack_start",
    #[cfg(not(feature = "single-hart"))]
    "sub t1, t1, t0",
    "andi sp, t1, -16 // align stack to 16-bytes
    add s0, sp, zero",
);
```

This level of assembler and bit fibbling is not my strong suite.

rmsyn: > But I'm unsure about the alignment. For example with 3 harts I get a stack size of 1362.

We can probably include an assert to ensure that the resulting stack size(s) satisfy alignment requirements.

> It looks like these instructions in _start might be there to fix that that though?

Yes, but as I understand it, the stack offsets will also need to be properly aligned. So, having the assert to check at the linking stage will still be useful.

romancardenas:
@janderholm , do you want to try the new approach of dividing equally the stack? And remember to include your changes to `CHANGELOG.md`

Issue #153, opened 2022-12-16
Linker relocation issue (QEMU / OpenSBI)


chutchinson:
I am experimenting with the RISC-V ISA via QEMU emulation, and I am having trouble figuring out how to link in the _heap_size symbol appropriately, so that I can follow through with heap allocation. I followed the **riscv-rt** documentation closely. I am attempting to use QEMU with OpenSBI, but I am not experienced with linker configuration. Here is my memory layout specification:

```text
MEMORY
{
    RAM : ORIGIN = 0x80200000, LENGTH = 0x8000000
    FLASH : ORIGIN = 0x20000000, LENGTH = 16M
}

REGION_ALIAS("REGION_TEXT", RAM);
REGION_ALIAS("REGION_RODATA", RAM);
REGION_ALIAS("REGION_DATA", RAM);
REGION_ALIAS("REGION_BSS", RAM);
REGION_ALIAS("REGION_HEAP", RAM);
REGION_ALIAS("REGION_STACK", RAM);
```

I understand that to mean that there is 16M of reserved flash memory at the beginning of the address space (OpenSBI), followed by RAM address space. I would expect all of the memory regions to fit within RAM since the FLASH space is reserved by QEMU OpenSBI firmware.

I have the following example program:

```rust
#![no_std]
#![no_main]

extern crate panic_halt;

mod uart;
mod print;

use riscv_rt::entry;

use uart::Uart;

extern "C" {
    static _heap_size: u8;
}

#[entry]
fn __start() -> ! {
    let heap_size = unsafe { &_heap_size as *const u8 as usize };
    println!("hello world");
    println!("heap size = {}", heap_size);
    loop {}
}
```

The problem is that the linker cannot relocate **_heap_size** provided by riscv-rt and I do not understand why. I tried changing the code model to **medium**:

```text
relocation R_RISCV_PCREL_HI20 out of range: -524800 is not in [-524288, 524287]; references _heap_size
```

vilukissa68:
We ran into this or a similar issue with @hegza and were able to solve it by setting `riscv64-unknown-elf-ld` as linker in `config.toml`
```
linker = "riscv64-unknown-elf-ld"
```

alistair23:
Why use OpenSBI though? Are you trying to write S-mode software?

hegza:
In our case it wasn't OpenSBI but an application thay needed access to a memory located in the high bits.

PR #297, opened 2025-06-11
Use la/jr pseudoinstructions for main function call

janderholm:
Replace direct jal with load address and jump register pseudoinstructions to support main functions located beyond the ±1MB range of direct jumps.

Together with a custom link script, this allows placing the bulk of the code in RAM which is likely located far away.

I use it for loading everything from a very slow SPI flash into SRAM on an FPGA. But it's useful for things like decreasing power consumption by powering off flash as well.

### DIFF
--- a/riscv-rt/src/asm.rs
+++ b/riscv-rt/src/asm.rs
@@ -217,7 +217,8 @@ cfg_global_asm!(
     ld a1, 8 * 1(sp)
     ld a2, 8 * 2(sp)
     addi sp, sp, 8 * 4",
-    "jal zero, main
+    "la t0, main
+    jr t0
     .cfi_endproc",
 );
 


romancardenas (approved):
LGTM! Maybe #153 can benefit from this PR.

@janderholm please add an entry into the `CHANGELOG.md` file

alistair23 (approved):

romancardenas:
> LGTM, too. Just out of curiosity, why is the separate quoting needed here (I see it was present before the changes, too)?

The code for RV32 and RV64 differs, but the call to main is the same

romancardenas:
@janderholm can you add an entry to `CHANGELOG.md` so this PR can be merged?

romancardenas (approved):

rmsyn (approved):

PR #299, opened 2025-06-13
Adapt `riscv-semihosting` to edition 2024

romancardenas:


romancardenas:
Looks like we need to update all the crates at once... I will keep doing this when I have time

riscv-semihosting/src/lib.rs
```diff
@@ -202,7 +202,7 @@ pub mod nr;
 /// [semihosting operation]: https://developer.arm.com/documentation/dui0471/i/semihosting/semihosting-operations?lang=en
 #[inline(always)]
 pub unsafe fn syscall<T>(nr: usize, arg: &T) -> usize {
-    syscall1(nr, arg as *const T as usize)
+    unsafe { syscall1(nr, arg as *const T as usize) }
```
rmsyn, line 205: Why is the `unsafe` block inside the `unsafe` function needed here?

riscv-semihosting/src/lib.rs
```diff
@@ -230,10 +231,11 @@ pub unsafe fn syscall1(_nr: usize, _arg: usize) -> usize {
...
             ",
-            inout("a0") nr,
-            inout("a1") arg => _,
-            options(nostack, preserves_flags),
-            );
+                inout("a0") nr,
+                inout("a1") arg => _,
+                options(nostack, preserves_flags),
+                )
+            };
```
rmsyn, line 238: Same as above, why is the `unsafe` block inside the `unsafe` function?

rmsyn:
> @rmsyn I think it is done. The only test failing is a nightly clippy lint that is still unstable.

Looks like I can't comment on the offending code, because you didn't make edits to it. Here is my suggestion:

```rust
    // riscv-rt/macros/src/lib.rs:91
    #[cfg(feature = "u-boot")]
    if let Some(argument) = f.sig.inputs.get(0) {
        if let Some(message) = check_argument_type(argument, "c_int") {
            return message;
        }
    }
    // instead do
    #[cfg(feature = "u-boot")]
    if let Some(message) = f.sig.inputs.get(0).and_then(|argument| check_argument_type(argument, "c_int")) {
        return message;
    }

    // same as above
    #[cfg(feature = "u-boot")]
    if let Some(argument) = f.sig.inputs.get(1) {
        if let Some(message) = check_argument_type(argument, "*const *const c_char") {
            return message;
        }
    }
    // should be
    #[cfg(feature = "u-boot")]
    if let Some(message) = f.sig.inputs.get(1).and_then(|argument| check_argument_type(argument, "*const *const c_char")) {
            return message;
    }
```

rmsyn (changes requested):
> @rmsyn I think it is done. The only test failing is a nightly clippy lint that is still unstable.

Looks like I can't comment on the offending code, because you didn't make edits to it. Here is my suggestion:

```rust
    // riscv-rt/macros/src/lib.rs:91
    #[cfg(feature = "u-boot")]
    if let Some(argument) = f.sig.inputs.get(0) {
        if let Some(message) = check_argument_type(argument, "c_int") {
            return message;
        }
    }
    // instead do
    #[cfg(feature = "u-boot")]
    if let Some(message) = f.sig.inputs.get(0).and_then(|argument| check_argument_type(argument, "c_int")) {
        return message;
    }

    // same as above
    #[cfg(feature = "u-boot")]
    if let Some(argument) = f.sig.inputs.get(1) {
        if let Some(message) = check_argument_type(argument, "*const *const c_char") {
            return message;
        }
    }
    // should be
    #[cfg(feature = "u-boot")]
    if let Some(message) = f.sig.inputs.get(1).and_then(|argument| check_argument_type(argument, "*const *const c_char")) {
            return message;
    }
```

Edit: here is a branch with formatted changes: https://github.com/rust-embedded/riscv/compare/edition-2024...rmsyn:riscv:edition-2024-lint-fixes?expand=1

rmsyn:
@romancardenas fixed the lints in #301

romancardenas:
Please, point to this branch in this PR and I will merge it. I think we should leave this PR updated and ready to merge as soon as we decide to go with edition 2024

rmsyn:
> as soon as we decide to go with edition 2024

After the discussion in today's meeting about updating `svd2rust` to edition 2024, I think the same arguments apply to the `riscv` crates, right? Since if `riscv` crates go to edition 2024, that would force `svd2rust` to update as well.

Linking the `svd2rust` discussion for tracking purposes: https://github.com/rust-embedded/svd2rust/pull/933#issuecomment-2954106460

PR #300, opened 2025-06-17
riscv-rt/macros: lint fixes for nightly

romancardenas:


romancardenas:
The proposed changes do not work as expected, it looks like message is consumed

PR #302, opened 2025-06-18
`riscv-peripheral`: const methods for HART 0

romancardenas:
Adds const methods to access registers for HART 0. This is especially handy for single-HART targets


rmsyn (approved):
LGTM.

PR #303, opened 2025-06-19
`riscv-semihosting`: remove custom build script

romancardenas:
As previously done in `riscv-rt`, this PR simplifies the build process of `riscv-semihosting`. Now, there is no need to define custom build flags.


rmsyn (approved):
LGTM.

PR #308, opened 2025-06-24
`riscv-rt`: validate hard ID earlier and allow `abort` to be located anywhere

romancardenas:
This PR moves the hart ID validation earlier in the boot process. Furthermore, as suggested by @janderholm , now `abort` can be located anywhere. This change may be especially important for targets with huge memory (see #153)


romancardenas:
Ooops, I used the wrong branch instruction. This new push fixes it. I have tested it with QEMU.

MabezDev (approved):

PR #313, opened 2025-06-26
`riscv`: add Xtvec::new and Xtvec::try_new

romancardenas:
Adds a few convenience functions to deal with `mtvec` and `xtvec` registers.


rmsyn (approved):
LGTM. We might also want to add builder-like methods in the future, but I think that is a separate conversation.

PR #317, opened 2025-06-29
RFC: Templated Linker Scripts

gibbz00:
This PR pulls in [`minilink`](https://github.com/gibbz00/minilink); a small wrapper around MiniJinja for linker scripts. I believe that this can make linker script customization far easier. For example, `esp-hal` places `.trap` sections in RAM (https://github.com/esp-rs/esp-hal/pull/541), and I would like to open a PR which does the same. Using templated linker scripts would make the addition fairly trivial:

```ld
{% if contains(cfg.feature, "trap-region") %}
.trap : ALIGN(4)
{
  *(.trap);
  *(.trap.*);
} > REGION_TRAP
{% endif %}

.text _stext :
{
  ...
  {% if not contains(cfg.feature, "trap-region") %}
  *(.trap);
  *(.trap.rust);
  {% endif %}
}
```

The biggest downside is the addition of build dependencies, and all that comes with it:

```
minilink v0.2.0
├── minijinja v2.11.0
│   └── serde v1.0.219
└── serde v1.0.219
```

Here are some `cargo build --timings --package riscv-rt` results for `{debug,release} x {warm,cold}` builds when executed on my laptop.

|              | master | templated | diff        |
| :--          | :--    | :--       | :--         |
| debug cold   | 2.1s   | 2.3s      | +0.2s (10%) |
| debug warm   | 0.2s   | 0.2s      | 0.0s (0%)   |
| release cold | 2.0s   | 2.2s      | +0.2s (10%) |
| release warm | 0.2s   | 0.2s      | 0.0s (0%)   |

Test varied with about ±0.05s when rerun. As you can see, warm build times remain pretty much unaffected. Cold builds had about 0.2s or 10% added to them. I personally think this is manageable given the advantages of templating with MiniJinja.

I chose to create a separate crate for this, as I believe that the linker script templating may be useful outside this repo. However, since `minilink` is about 200 lines of code—documentation excluded—there's not much in the way of copying relevant parts directly into `riscv-rt`.

### Alternatives

`esp-hal` has a [very simple C-like preprocessor](https://github.com/esp-rs/esp-hal/blob/9d452a108fcce7cf0759b125449234c5e4e4c190/esp-hal/build.rs#L159-L160). Suppose something similar could be used. But there is something to be said about using a less bespoke, and well documented solution.

### DIFF
--- a/tests-build/Cargo.toml
+++ b/tests-build/Cargo.toml
@@ -8,6 +8,9 @@ panic-halt = "1.0"
 riscv = { path = "../riscv" }
 riscv-rt = { path = "../riscv-rt" }
 
+[build-dependencies]
+minilink = "0.2"
+
 [features]
 pre-init = ["riscv-rt/pre-init"]
 single-hart = ["riscv-rt/single-hart"]
--- a/tests-build/build.rs
+++ b/tests-build/build.rs
@@ -1,22 +1,6 @@
 use std::{env, fs::File, io::Write, path::PathBuf};
 
 fn main() {
-    // Put device.x somewhere the linker can find it
-    let out = &PathBuf::from(env::var_os("OUT_DIR").unwrap());
-    File::create(out.join("device.x"))
-        .unwrap()
-        .write_all(include_bytes!("device.x"))
-        .unwrap();
-    println!("cargo:rustc-link-search={}", out.display());
-    println!("cargo:rerun-if-changed=device.x");
-
-    // Put memory.x somewhere the linker can find it
-    File::create(out.join("memory.x"))
-        .unwrap()
-        .write_all(include_bytes!("memory.x"))
-        .unwrap();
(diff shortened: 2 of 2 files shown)


rmsyn (commented):
I personally don't see a need to bring in a new dependency for our use-case in `riscv-rt`. The build script handles the `cfg` dependencies fairly well (although, the `minilink`-based config does look cleaner). If more of the linker script contains `cfg`-dependent sections, I could see more justification for the additional dependency.

Not strongly opposed to the idea, just offering my opinion.

romancardenas:
I like the idea! Let's keep this PR as a draft and then we can decide whether it is worth it.

I will probably open a few PRs that tidy up the linker file a bit, so they might affect this draft. The idea is approaching slowly to a state in which `esp-riscv-rt` can be built on top of `riscv-rt`, so probably people from `esp-rs` don't want to give up linking features.

PR #318, opened 2025-06-30
docs(riscv-rt): make feature flag descriptions more succinct

gibbz00:
* Avoids referring to how cargo features work with an example of how to use them for every feature mentioned.
* Mention that no features flags are enabled by default.
* Also: Remove outdated abort comment in `link.x.in`. I couldn't be bothered to open a separate PR for this, hope that's ok.


romancardenas (approved):
LGTM!

PR #322, opened 2025-07-01
Join global_asm macro invocations

### LINKED ISSUE #321: asm.rs assumes that multiple global_asm statements are concatenated
In https://github.com/rust-embedded/riscv/blob/master/riscv-rt/src/asm.rs there are multiple `global_asm!` macro calls (generated from multiple `cfg_global_asm!` macro calls`) with code that only make sense if the generated machine code is a single contiguous block.

However, as far as I can tell, rust doesn't guarantee that.

To the contrary, https://dirname.github.io/rust-std-doc/unstable-book/library-features/global-asm.html contains the following language: "However, you should not rely on assembler state (e.g. assembler macros) defined in one global_asm! to be available in another one. It is implementation-defined whether the multiple usages are concatenated into one or assembled separately."

(Of course, even if assembled separately, the generated code could still be contiguous. I just didn't find any guarantee that it will.)

In the rust-embedded matrix channel, @Dirbaio suggested to use a single `cfg_global_asm!` block. To avoid unwieldy repetitive `cfg` statements, the `cfg_global_asm!` macro could be extended to allow for nested blocks:
```

/// Parse cfg attributes inside a global_asm call.
macro_rules! cfg_global_asm {
    {@inner, [$($x:tt)*], } => {
        global_asm!{$($x)*}
    };
    (@inner, [$($x:tt)*], #[cfg($meta:meta)] { $($y:tt)* } $($rest:tt)*) => {
        #[cfg($meta)]
        cfg_global_asm!{@inner, [$($x)*], $($y)* $($rest)*}
        #[cfg(not($meta))]
        cfg_global_asm!{@inner, [$($x)*], $($rest)*}
    };
    (@inner, [$($x:tt)*], #[cfg($meta:meta)] $asm:literal, $($rest:tt)*) => {
        #[cfg($meta)]
        cfg_global_asm!{@inner, [$($x)* $asm,], $($rest)*}
        #[cfg(not($meta))]
        cfg_global_asm!{@inner, [$($x)*], $($rest)*}
    };
    {@inner, [$($x:tt)*], $asm:literal, $($rest:tt)*} => {
        cfg_global_asm!{@inner, [$($x)* $asm,], $($rest)*}
    };
    {$($asms:tt)*} => {
        cfg_global_asm!{@inner, [], $($asms)*}
    };
}
```
untested, could be used like this:
```
cfg_global_asm!(
    "blah",
    #[cfg(foo)]
    "blah",
    #[cfg(bar)] {
        "blah",
        #[cfg(foo)]
        "blah",
    }
)
```
(https://libera.catirclogs.org/rust-embedded/2025-07-01#38366632)

jannic:
Fixes #321

I intentionally didn't update the indentation, for two reasons:
1) I don't even know how this should be indented
2) This way it's easier to review

If you'd prefer some certain way of indenting the conditional blocks inside the `cfg_global_asm!` call, I'll happily update the pull request.

### DIFF
--- a/riscv-rt/CHANGELOG.md
+++ b/riscv-rt/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - `clippy` fixes
+- Merged `cfg_global_asm!` macro invocations to guarantee contiguous code generation.
 
 ## [v0.15.0] - 2025-06-10
 
--- a/riscv-rt/src/asm.rs
+++ b/riscv-rt/src/asm.rs
@@ -5,6 +5,12 @@ macro_rules! cfg_global_asm {
     {@inner, [$($x:tt)*], } => {
         global_asm!{$($x)*}
     };
+    (@inner, [$($x:tt)*], #[cfg($meta:meta)] { $($y:tt)* } $($rest:tt)*) => {
+        #[cfg($meta)]
+        cfg_global_asm!{@inner, [$($x)*], $($y)* $($rest)*}
+        #[cfg(not($meta))]
+        cfg_global_asm!{@inner, [$($x)*], $($rest)*}
+    };
     (@inner, [$($x:tt)*], #[cfg($meta:meta)] $asm:literal, $($rest:tt)*) => {
         #[cfg($meta)]
         cfg_global_asm!{@inner, [$($x)* $asm,], $($rest)*}
@@ -80,17 +86,14 @@ _abs_start:
     la t0, abort // If hart_id > _max_hart_id, jump to abort
     jr t0
 1:", // only valid harts reach this point
-);
 
 // INITIALIZE GLOBAL POINTER, STACK POINTER, AND FRAME POINTER
-cfg_global_asm!(
     ".option push
     .option norelax
     la gp, __global_pointer$
     .option pop",
-);
 #[cfg(not(feature = "single-hart"))]
-cfg_global_asm!(
+{
     "mv t2, a0
     lui t0, %hi(_hart_stack_size)
     add t0, t0, %lo(_hart_stack_size)",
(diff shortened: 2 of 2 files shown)


rmsyn (changes requested):
Minor clippy fix:

https://github.com/rust-embedded/riscv/actions/runs/16010020609/job/45165521521?pr=322

Otherwise, LGTM.

romancardenas:
> Minor clippy fix:
> 
> https://github.com/rust-embedded/riscv/actions/runs/16010020609/job/45165521521?pr=322
> 
> Otherwise, LGTM.

`is_multiple_of` came with Rust 1.87: https://doc.rust-lang.org/std/primitive.usize.html#method.is_multiple_of

Bumping the MSRV to make Nightly Clippy happy seems too much. Currently, Nightly builds and checks are marked as experimental, and it is OK to merge PRs with reasonable failures in Nightly.

I'm CC'ing @gibbz00 , as he mentioned that it might be possible to "relax" a bit Nightly clippy requirements

romancardenas (approved):
LGTM! Thank you.

There are a few `core::arch::global_asm!` calls in the macros and at the end of `asm.rs`, but all of these blobs specify a section, so I guess it is not UB? We don't need them to be in a specific memory location or order, as long as they are in their corresponding section.

gibbz00:
> I'm CC'ing @gibbz00 , as he mentioned that it might be possible to "relax" a bit Nightly clippy requirements

`clippy::manual_is_multiple_of` is unfortunately part of the "complexity" lint group, whose default lint level is "warn". I would suggest adding an `#[allow(clippy::manual_is_multiple_of, reason = "requires MSRV bump")]` for now. Relaxing clippy lints would entail removing this line: https://github.com/rust-embedded/riscv/blob/cdae64a64a5a22d5e8fb0a2ffaa98dc60ccb80cd/.github/workflows/clippy.yaml#L10

(Concise lint group explanation here: https://github.com/rust-lang/rust-clippy?tab=readme-ov-file#clippy)

That is, the general lint policy should IMO be;

* Avoid opting-out of default lints unless there's a strong reason for it. (Like the one above.)
* Selectively opt-in to non-default/allow-by-default lints at the lint level, not lint group level. Ex. by adding `-W missing-docs`.

jannic:
Thanks for discussing the clippy warnings. I didn't have time to care about them yesterday. I agree that it doesn't make sense chasing every single new warning on nightly immediately.

PR #324, opened 2025-07-03
`riscv-rt`: Use `_start_rust` and `hal_main`

romancardenas:
This PR goes back to using `_start_rust`, reducing the assembly code to the minimum. Also, inspired by `esp-riscv-rt`, now `_start_rust` jumps to `hal_main` instead of `main`. At linker level, if `hal_main` is not defined, it maps to `main`, so it would still work in the same was as of today.


MabezDev (approved):
Makes sense to me, thanks!

PR #326, opened 2025-07-04
`riscv-rt`: Preserve `a0-a2` in startup only when startup functions are expected

### LINKED ISSUE #315: RAM is used before __pre_init.
The following statement is found in the changelog for v0.15.0:

```
New pre-init feature to run a __pre_init function before RAM initialization.
```

This makes sense, since clocks, timings and other stuff might need to be initialized for some memory to work. However, to me it looks like the stack is used prior to the call to `__pre_init`:

```rust
// STORE A0..A2 IN THE STACK, AS THEY WILL BE NEEDED LATER BY main
cfg_global_asm!(
    "addi sp, sp, -4 * 4 // we must keep stack aligned to 16-bytes
    sw a0, 4 * 0(sp)
    sw a1, 4 * 1(sp)
    ...
```

followed by :
```rust
// CALL __pre_init (IF ENABLED) AND INITIALIZE RAM
cfg_global_asm!(
    #[cfg(not(feature = "single-hart"))]
    // Skip RAM initialization if current hart is not the boot hart
    "call _mp_hook
    beqz a0, 4f",
    #[cfg(feature = "pre-init")]
    "call __pre_init",
    "// Copy .data from flash to RAM
    ....
```

Is this an oversight, or do I miss some key piece of the puzzle? If the stack is located in RAM that needs to be initialized it might get corrupted.

I don't currently have this problem, I'm just scratching my head over it a bit.

romancardenas:
Alternative to #325 . This is simpler, as it reduces register preservation to the minimum. New documentation specifies that registers `a1` and `a2` must not be modified during `_mp_hook` and `__pre_init`.

I personally prefer this over #325, but let me know your opinion.

Closes #315

### DIFF
--- a/riscv-rt/CHANGELOG.md
+++ b/riscv-rt/CHANGELOG.md
@@ -31,6 +31,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Now, `_start_rust` jumps to `hal_main` instead of `main` directly. At linker level,
   `hal_main` maps to `main` if not defined. However, we now allow HALs to inject
   additional configuration code before jumping to the final user's `main` function.
+- Now, `a0-a2` are preserved in `s0-s2` during the startup process. In RVE targets,
+  `a2` is preserved in `a5`, as there are only two callee-saved registers.
+  New documentation of startup functions (`_mp_hook` and `__pre_init`) now provide
+  additional implementation guidelines to ensure a correct behavior of the runtime.
+
+### Removed
+
+- Removed usage of the stack before `_start_rust`. This was unsound, as in the `.init`
+  section RAM is still uninitialized.
 
 ### Fixed
 
--- a/riscv-rt/src/asm.rs
+++ b/riscv-rt/src/asm.rs
@@ -85,9 +85,18 @@ _abs_start:
     bgeu t0, a0, 1f
     la t0, abort // If hart_id > _max_hart_id, jump to abort
     jr t0
-1:", // only valid harts reach this point
-
-// INITIALIZE GLOBAL POINTER, STACK POINTER, AND FRAME POINTER
+1:", // Only valid hart IDs reach this point
+#[cfg(any(feature = "pre-init", not(feature = "single-hart")))]
+// If startup functions are expected, preserve a0-a2 in s0-s2
+{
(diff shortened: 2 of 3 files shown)


riscv-rt/CHANGELOG.md
```diff
@@ -31,6 +31,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Now, `_start_rust` jumps to `hal_main` instead of `main` directly. At linker level,
   `hal_main` maps to `main` if not defined. However, we now allow HALs to inject
   additional configuration code before jumping to the final user's `main` function.
+- Now, `a0` is preserved in `s1` during the startup process.
+
+### Removed
+
+- Removed usage of the stack before `_start_rust`. This was unsound, as in the `.init`
+  section RAM is still uninitialized.
+- Now, `a1` and `a2` are **not** preserved during the startup process. New documentation
```
rmsyn, line 40: I don't think clobbering `a1` at startup is a good idea. For instance, if `riscv-rt` code is being run as a kernel-like SBI payload, `a1` stores the address location of the `DeviceTree`.

I think this is just a convention, so not a hard requirement. I'm not aware of any conventions requiring preservation of the `a2` register.

romancardenas: In RVI, we can store `a0-a2` in `s1-s3`. We can even use `s0-s2` and initialize the frame pointer later, as proposed in the other PR. But im RVE we only have `s0-s1`, so we need to use `a4` and/or `a5`. Using `aX` registers from preservation is not part of the ABI, so we must document it to avoid misuse.

Here, while we are not preserving `a1-a2`, we clearly specify in the documentation for `_mp_hook` and `__preinit` that these registers are reserved and must not be used.

Let me know what you prefer and I'll try it.

romancardenas:
@hegza can you check this PR too? Which one do you prefer?

rmsyn (approved):

hegza:
Still compiles on RVE. I have no preference whatsoever, both work. One nice thing about this is that the only implementation that gets a bit of extra complexity is the mythical multicore RVE. I'm pretty sure that if anyone's going to build one anytime soon it's going to be our lab 😅

PR #328, opened 2025-07-08
`riscv-rt`: Add `__post_init`  Rust function

romancardenas:
Inspired by `esp-riscv-rt`. It includes a `__post_init` Rust function to perform initialization actions safely from Rust.

The new `__post_init` function is optional and feature-gated. If enabled, no default symbol is provided, and users must include this function. They can use the new `#[riscv_rt::post_init]` attribute macro. I included `trybuild` tests.

### DIFF
--- a/riscv-rt/macros/src/lib.rs
+++ b/riscv-rt/macros/src/lib.rs
@@ -63,26 +63,6 @@ pub fn entry(args: TokenStream, input: TokenStream) -> TokenStream {
         .into();
     }
 
-    fn check_correct_type(argument: &PatType, ty: &str) -> Option<TokenStream> {
-        let inv_type_message = format!("argument type must be {ty}");
-
-        if !is_correct_type(&argument.ty, ty) {
-            let error = parse::Error::new(argument.ty.span(), inv_type_message);
-
-            Some(error.to_compile_error().into())
-        } else {
-            None
-        }
-    }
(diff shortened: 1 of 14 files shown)


riscv-rt/macros/src/lib.rs
```diff
@@ -263,6 +265,93 @@ pub fn pre_init(args: TokenStream, input: TokenStream) -> TokenStream {
...
+        .into();
+    }
+
+    if !args.is_empty() {
+        return parse::Error::new(Span::call_site(), "This attribute accepts no arguments")
+            .to_compile_error()
+            .into();
+    }
+
+    // XXX should we blacklist other attributes?
```
rmsyn, line 341: Are there any other attributes that could make the function unsound? We could also require that the macro is `unsafe`, and document the invariants.

rmsyn (approved):
Small question, but LGTM overall.

Thanks for adding the `trybuild` tests!

riscv-rt/macros/src/lib.rs
romancardenas: I mean, `__post_init` is meant to be executed in the Rust-safe world, so there is not much trouble with it.

rmsyn: Sounds good, if there are no attributes that could make the implementation unsound. Thanks for the explanation.

PR #330, opened 2025-07-15
`riscv`: Use `riscv_pac::CoreInterrupt` in `mie` register

### LINKED ISSUE #314: `riscv`: Enable setting external interrupt bits in mie
The low power core in the Espressif esp32c6 microcontroller has all external interrupts routed to interrupt 30, but the current mask on mie prevents the corresponding bit from being set.

romancardenas:
This PR adds new methods and functions for the `mie` CSR. The idea is to use the `riscv_pac::CoreInterruptNumber` trait to allow enabling and disabling target-specific interrupt sources (i.e., extending our API to non-standard interrupt sources).

Closes #314

### DIFF
--- a/riscv/src/register/mie.rs
+++ b/riscv/src/register/mie.rs
@@ -89,7 +89,7 @@ set_clear_csr!(
 
 /// Disables a specific core interrupt source.
 #[inline]
-pub fn disable_interrupt<I: CoreInterruptNumber>(interrupt: I) {
+pub fn disable<I: CoreInterruptNumber>(interrupt: I) {
     // SAFETY: it is safe to disable an interrupt source
     unsafe { _clear(1 << interrupt.number()) };
 }
@@ -101,7 +101,7 @@ pub fn disable_interrupt<I: CoreInterruptNumber>(interrupt: I) {
 /// Enabling interrupts might break critical sections or other synchronization mechanisms.
 /// Ensure that this is called in a safe context where interrupts can be enabled.
 #[inline]
-pub unsafe fn enable_interrupt<I: CoreInterruptNumber>(interrupt: I) {
+pub unsafe fn enable<I: CoreInterruptNumber>(interrupt: I) {
     unsafe { _set(1 << interrupt.number()) };
 }
 
--- a/riscv/src/register/sie.rs
+++ b/riscv/src/register/sie.rs
@@ -62,7 +62,7 @@ set_clear_csr!(
 
 /// Disables a specific core interrupt source.
 #[inline]
-pub fn disable_interrupt<I: CoreInterruptNumber>(interrupt: I) {
+pub fn disable<I: CoreInterruptNumber>(interrupt: I) {
     // SAFETY: it is safe to disable an interrupt source
     unsafe { _clear(1 << interrupt.number()) };
 }
@@ -74,7 +74,7 @@ pub fn disable_interrupt<I: CoreInterruptNumber>(interrupt: I) {
 /// Enabling interrupts might break critical sections or other synchronization mechanisms.
 /// Ensure that this is called in a safe context where interrupts can be enabled.
 #[inline]
-pub unsafe fn enable_interrupt<I: CoreInterruptNumber>(interrupt: I) {
+pub unsafe fn enable<I: CoreInterruptNumber>(interrupt: I) {
     unsafe { _set(1 << interrupt.number()) };
 }
 
--- a/riscv/CHANGELOG.md
+++ b/riscv/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - New convenience  `try_new` and `new` associated functions for `Mtvec` and `Stvec`.
 - New methods and functions for enabling core interrupts in the `mie` and `sie` registers
   using the `riscv_pac::CoreInterrupt` trait.
-- New `riscv::interrupt::{disable_interrupt, enable_interrupt}` functions.
+- New `riscv::interrupt::{is_interrupt_enabled, disable_interrupt, enable_interrupt}` functions.
 
 ### Changed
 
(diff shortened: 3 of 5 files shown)


riscv/src/register/mie.rs
```diff
@@ -42,6 +44,26 @@ read_write_csr_field! {
     mext: 11,
 }
 
+impl Mie {
+    /// Check if a specific core interrupt source is enabled.
+    #[inline]
+    pub fn is_enabled<I: CoreInterruptNumber>(&self, interrupt: I) -> bool {
+        (self.bits & (1 << interrupt.number())) != 0
```
rmsyn, line 51:
```suggestion
        bf_extract(self.bits, interrupt.number(), 1) != 0
```

romancardenas:
Ooops! I pushed a few more changes. Can you please review the PR again? I will address your comments after your new review!

riscv/src/register/sie.rs
```diff
@@ -1,9 +1,11 @@
 //! sie register
 
+use riscv_pac::CoreInterruptNumber;
```
rmsyn, line 3:
```suggestion
use riscv_pac::CoreInterruptNumber;
use crate::bits::{bf_insert, bf_extract};
```

riscv/src/register/sie.rs
```diff
@@ -24,6 +26,26 @@ read_write_csr_field! {
     sext: 9,
 }
 
+impl Sie {
+    /// Check if a specific core interrupt source is enabled.
+    #[inline]
+    pub fn is_enabled<I: CoreInterruptNumber>(&self, interrupt: I) -> bool {
+        (self.bits & (1 << interrupt.number())) != 0
```
rmsyn, line 33:
```suggestion
        bf_extract(self.bits, interrupt.number(), 1) != 0
```

riscv/src/register/sie.rs
```diff
@@ -24,6 +26,26 @@ read_write_csr_field! {
...
+    }
+
+    /// Enable a specific core interrupt source.
+    #[inline]
+    pub fn enable<I: CoreInterruptNumber>(&mut self, interrupt: I) {
+        self.bits |= 1 << interrupt.number();
```
rmsyn, line 39:
```suggestion
        self.bits = bf_insert(self.bits, interrupt.number(), 1, 1);
```

riscv/src/register/sie.rs
```diff
@@ -24,6 +26,26 @@ read_write_csr_field! {
...
+    }
+
+    /// Disable a specific core interrupt source.
+    #[inline]
+    pub fn disable<I: CoreInterruptNumber>(&mut self, interrupt: I) {
+        self.bits &= !(1 << interrupt.number());
```
rmsyn, line 45:
```suggestion
        self.bits = bf_insert(self.bits, interrupt.number(), 1, 0);
```

riscv/src/register/mie.rs
```diff
@@ -100,7 +100,7 @@ pub fn clear_interrupt<I: CoreInterruptNumber>(interrupt: I) {
 /// Enabling interrupts might break critical sections or other synchronization mechanisms.
 /// Ensure that this is called in a safe context where interrupts can be enabled.
 #[inline]
-pub unsafe fn set_interrupt<I: CoreInterruptNumber>(interrupt: I) {
+pub unsafe fn enable_interrupt<I: CoreInterruptNumber>(interrupt: I) {
```
rmsyn, line 103: Is there a reason you're not including the `is_enabled`, `enable`, `disable` methods for the `Mie` type, but are including them for the `Sie` type?

rmsyn (changes requested):
Small change requests for code-reuse, and type methods. Otherwise, LGTM.

riscv/src/register/mie.rs
rmsyn: Sorry, the changes just didn't show up in the diff I was looking at, never mind. :sweat_smile:

rmsyn (approved):

romancardenas:
Before merging, I have some doubts about the naming:

What do you think about having `fn riscv::register::mie::enable<I: CoreInterruptNumber>()` and fn riscv::interrupt::enable_interrupt<I: CoreInterruptNumber>()`? Or do you prefer both functions to have the same name?

Additionally, should I include `riscv::interrupt::is_interrupt_enabled<I>()`? Maybe we don't want this kind of functions exposed in `riscv::interrupt`

rmsyn:
> What do you think about having fn riscv::register::mie::enable<I: CoreInterruptNumber>() and fn riscv::interrupt::enable_interrupt<I: CoreInterruptNumber>()`?

Either one is fine, I don't have strong opinions either way. `enable` is shorter, `enable_interrupt` is more explicit (though maybe redundant).

> Or do you prefer both functions to have the same name?

Whatever name you choose, I would personally prefer consistent naming. Up to you, though.

> Additionally, should I include riscv::interrupt::is_interrupt_enabled<I>()? Maybe we don't want this kind of functions exposed in riscv::interrupt

I think it would be good to include it. The same type of user that would want the other free functions would probably also prefer this one to be available.

rmsyn:
> Should mie::enable be unsafe?

I think, for the same reasons enabling `mie` in `mstatus` is unsafe, enabling individual interrupt sources in `mie` should also be unsafe.

We have no way of knowing that all interrupt sources on all platforms are safe to enable. We also know that some interrupt sources are unsafe/unsound to enable in some contexts.

In my opinion, the above means that we should mark it / leave it marked unsafe, and describe the invariants users need to consider / maintain. Another point is that we cannot, without a very convoluted setup, determine `mstatus::mie` settings at runtime to know when it is safe vs. unsafe to enable interrupt sources in the `mie` CSR. This would probably be better left to higher-level libraries to provide those kind of abstractions.

rmsyn (approved):

PR #334, opened 2025-07-24
Allow 64-bit Mcause::from_bits on rv64

### LINKED ISSUE #333: Mcause::from_bits doesn't allow interrupt to be set.
`Mcause::from_bits` is documented as only setting `BITMASK` bits, but this is set to `0xffff_ffff`. This excludes the interrupt bit on rv64, at bit 63. It should include at least this bit, but probably all 64 bits, since the riscv privileged spec says the Exception Code field is MXLEN-1 (with Interrupt at MXLEN-1)

jsgf:
Set the mcause mask to `!0u64` on rv64 targets.

Fixes #333

### DIFF
--- a/riscv/src/register/mcause.rs
+++ b/riscv/src/register/mcause.rs
@@ -5,7 +5,7 @@ pub use crate::interrupt::Trap;
 read_only_csr! {
     /// `mcause` register
     Mcause: 0x342,
-    mask: 0xffff_ffff,
+    mask: usize::MAX,
 }
 
 #[cfg(target_arch = "riscv32")]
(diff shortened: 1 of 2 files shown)


riscv/src/register/mcause.rs
```diff
@@ -5,7 +5,7 @@ pub use crate::interrupt::Trap;
 read_only_csr! {
     /// `mcause` register
     Mcause: 0x342,
-    mask: 0xffff_ffff,
+    mask: !0usize,
```
romancardenas, line 8:
```suggestion
    mask: usize::MAX,
```

romancardenas (changes requested):
Thanks! In newer PRs, we are using `usize::MAX` instead of `!0usize`. Can you change this?

PR #335, opened 2025-07-31
Optionally implement `defmt::Format`

bjoernQ:
Not a huge thing but since `Debug` is already implemented on these two types, having `defmt::Format` would be useful (at least for us and at least for `TrapFrame` ).

Given `defmt` is 1.0 this shouldn't be a problem, I guess. Let me know if you want something changed here or if you in general dislike the optional dependency.


romancardenas (approved):
LGTM! `defmt` is a very typical dependency and feature, so I guess it is OK to merge this PR. Thank you!

romancardenas:
Looks amazing!

You can use the `#[riscv::core_interrupt]` macro over your enumeration of interrupt sources. The macro is supposedly able to create the vector table and the default `_start_INTERRUPT_trap` symbols. I will take a look and find if we need to adapt something.

Issue #226, opened 2024-08-19
Advanced Interrupt Architecture support


bjorn3:
Among other things this defines a couple of CSR's for the IMSIC:

* [ ] `miselect`
* [ ] `mireg`
* [ ] `mtopei`
* [ ] `mtopi`
* [ ] `mvien`
* [ ] `mvip`
* [ ] `siselect`
* [ ] `sireg`
* [ ] `stopei`
* [ ] `stopi`
* [ ] `hvien`
* [ ] `hvictl`
* [ ] `hviprio1`
* [ ] `hviprio2`
* [ ] `vsiselect`
* [ ] `vsireg`
* [ ] `vstopei`
* [ ] `vstopi`

PR #343, opened 2025-09-11
riscv: add the `miselect` CSR

0xllx0:
Adds the `miselect` register for selecting an indirect CSR according to the `Smcsrind` extension.

Related: #1, #226


romancardenas (approved):
LGTM!

PR #346, opened 2025-09-22
Add `rt` feature gate to riscv::pac_enum

romancardenas:
An alternative to #345 

It presents way fewer changes, but now we expect caller crates of the `pac_enum` macro to include two features:

- `rt`
- `v-trap`

I personally prefer this approach, but let me know your opinion.

Related issues: https://github.com/rust-embedded/svd2rust/issues/948

### DIFF
--- a/riscv/CHANGELOG.md
+++ b/riscv/CHANGELOG.md
@@ -12,6 +12,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Add `miselect` CSR
 - Improved assembly macro handling in asm.rs
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


rmsyn:
> There is a drawback with this approach: PACs must include both rt and v-trap features, even if the target does not support vectored mode at all.

I think your approach in #345 is probably preferable. Especially since the macros live in that crate, it makes sense to add the features there, rather than in the generated code.

It will probably also be easier for downstream users to understand, rather than magic new features that show up in their code, and having to trace the source back to the proc macro.

romancardenas:
I agree with you @rmsyn . Adding magical feature gates in proc macros does not sound like a good practice. Closing in favor of #345

PR #349, opened 2025-09-28
Fix: Don't crash when target-feature=+relax is given

wisp3rwind:
This feature is a bit unusual in that it does not refer to actual hardware features, but is rather a target-specific linker feature. This PR simply skips this feature when building `RiscvTarget`, since from my understanding, it is irrelevant here.

See also https://github.com/rust-lang/rust/pull/109860

### DIFF
--- a/riscv-target-parser/src/lib.rs
+++ b/riscv-target-parser/src/lib.rs
@@ -106,6 +106,12 @@ pub struct RiscvTarget {
     extensions: Extensions,
 }
 
+// Returns whether a target feature _might_ be an ISA extension according to a non-exhaustive list
+// of known unrelated features flags.
+fn is_isa_extension(feature: &str) -> bool {
+    feature != "relax"
+}
+
 impl RiscvTarget {
     /// Builds a RISC-V target from a target triple and cargo flags.
     /// This function is expected to be called from a build script.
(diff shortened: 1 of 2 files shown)


riscv-target-parser/src/lib.rs
```diff
@@ -135,11 +143,15 @@ impl RiscvTarget {
             })
         {
             if let Some(feature) = target_feature.strip_prefix('+') {
-                let extension = Extension::try_from(feature)?;
-                target.extensions.insert(extension);
+                if !not_an_isa_extension(feature) {
```
romancardenas, line 146: To avoid double negative clauses, I would suggest you to change the function to `is_isa_extension` instead of `not_an_isa_extension`:

```suggestion
                if is_isa_extension(feature) {
```

romancardenas (changes requested):
Thanks for your contribution!

Apart from my comment about double negative clauses, please, address the following issues:

- Add a small entry in `riscv-target-parser/CHANGELOG.md` to describe the changes.
- Make all the required changes to make clippy happy: https://github.com/rust-embedded/riscv/actions/runs/18076762062/job/51463063101?pr=349

riscv-target-parser/src/lib.rs
wisp3rwind: The idea with this naming was that the check in the new helper function is not exhaustive, `!not_an_isa_extension(f)` doesn't really imply that `f` is an extension. New feature flags like "relax" could be added.

That said, I agree that this is hard to read. I'm applying the change you suggest and will extend the code comment about `is_isa_extension` to mention this.

romancardenas (approved):
LGTM

PR #347, opened 2025-09-24
Add mtopi CSR

KushalMeghani1644:
This PR adds support for the `mtopi` (Machine Top Priority Interrupt) CSR, which is part of the RISC-V Advanced Interrupt Architecture (AIA) extension.

**Changes:**
- Add `mtopi.rs` register implementation at address 0x7C0
- Export mtopi module in `register.rs`
- Read-only CSR with proper bit field definitions:
  - `iid` (Interrupt ID): bits 16-27 
  - `ipid` (Interrupt Priority ID): bits 0-7
- Convenience methods for interrupt detection and priority handling

**Features:**
- Memory-safe read-only access
- Comprehensive documentation with usage examples
- Full test coverage
- Follows existing CSR implementation patterns

**Testing:**
- All existing tests continue to pass
- New tests cover bit field extraction and edge cases
- Documentation examples compile successfully

**AI Assistance:**
Parts of the explanation and understanding of the mtopi CSR were assisted using an AI tool to summarize its behavior in the context of the RISC-V Advanced Interrupt Architecture (AIA). All code and implementation decisions were written and verified manually. NO CODES WERE GENERATED WITH THE HELP OF AI.

PR #357, opened 2025-10-29
`riscv`: Run tests on CI

romancardenas:
While reviewing #347 , I noticed that a test should have failed but CI didn't show any issue. It looks like our current CI does not run `riscv` tests, as it only checks that it can compile. This PR should fix this.


MabezDev (approved):
Thanks!

PR #358, opened 2025-10-30
Move `riscv-macros` to root directory

romancardenas:
The idea is trying to sanitize the procedural macros crates that we have (namely, `riscv-macros` and `riscv-rt-macros`). Now that `riscv-macros` has a `rt` feature, I think it would make sense to merge both crates and try to reuse utility parts such as all the code that adapts to the final target.

### DIFF
--- a/riscv-macros/src/lib.rs
+++ b/riscv-macros/src/lib.rs
@@ -437,16 +437,14 @@ core::arch::global_asm!("
 ///     E3 = 3,
 /// }
 ///
-/// fn main() {
-///     assert_eq!(Exception::E1.number(), 1);
-///     assert_eq!(Exception::E3.number(), 3);
+/// assert_eq!(Exception::E1.number(), 1);
+/// assert_eq!(Exception::E3.number(), 3);
 ///
-///     assert_eq!(Exception::from_number(1), Ok(Exception::E1));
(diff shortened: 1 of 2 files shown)


rmsyn (commented):
> The idea is trying to sanitize the procedural macros crates that we have (namely, riscv-macros and riscv-rt-macros). Now that riscv-macros has a rt feature, I think it would make sense to merge both crates and try to reuse utility parts such as all the code that adapts to the final target.

It looks like you've only moved over the `riscv/macros` so far. Are you planning to bring in the `riscv-rt/macros` stuff as well? Maybe feature-gated behind the `rt` feature?

For the stuff brought over from `riscv/macros` it might also make sense to put it behind a `#[cfg(not(feature = "rt"))]` (or introduce a new feature), to avoid bringing in stuff that the `riscv-rt` crate doesn't use. Not sure how it would affect things, since these are proc-macros, and I'm assuming what doesn't get used doesn't get compiled in. So, maybe a bit of overengineering.

romancardenas:
@rmsyn yes, in a set of future PRs I will try to merge both crates wisely. Regarding things in `riscv-macros` that are not used by `riscv-rt`, `riscv-rt` depends on `riscv`, so I guess it shouldn't be a problem.

PR #360, opened 2025-10-31
Add mtopi CSR

KushalMeghani1644:
This PR reintroduces the mtopi CSR addition from my previous PRs (#347  and #359), 
but from a clean branch as suggested by the reviewers. 
The diff is identical to the intended changes; commit history has been cleaned up.

### DIFF
--- a/riscv/src/register/mtopi.rs
+++ b/riscv/src/register/mtopi.rs
@@ -62,7 +62,6 @@ mod tests {
     #[test]
     fn test_mtopi_fields() {
         let mtopi = Mtopi::from_bits(0);
-
         test_ro_csr_field!(mtopi, iid: [16, 27], 0x0);
         test_ro_csr_field!(mtopi, iprio: [0, 7], 0x0);
 


romancardenas (changes requested):
Thanks! Just a few changes: add the field check in the new macro to check the field bit range and use hex representation in a test.

riscv/src/register/macros.rs
```diff
@@ -1070,3 +1070,19 @@ macro_rules! test_csr_field {
...
 }
+
+#[cfg(test)]
+#[macro_export]
+macro_rules! test_ro_csr_field {
+    ($reg:ident, $field:ident: [$start:expr, $end:expr], $expected:expr) => {{
+        let bits = $reg.bits();
+        let shift = $end - $start + 1;
+        let mask = (1usize << shift) - 1;
+        let exp_val = (bits >> $start) & mask;
```
romancardenas:
```suggestion
        let exp_val = $crate::bits::bf_extract(bits, $start, $end - $start + 1);
```

romancardenas (changes requested):
Thanks! Minor changes

KushalMeghani1644:
Thanks! @romancardenas for giving the review + commits, I have applied them, if there's any other change I'll happily make them :)

romancardenas (approved):
LGTM

PR #361, opened 2025-10-31
riscv: add the `mvien` + `mvienh` CSR

ghost:
Adds the `mvien` + `mvienh` CSR to represent the `Machine Virtual Interrupt Enable` registers.

Related: #1, #226

### DIFF
--- a/riscv/src/register/mvien.rs
+++ b/riscv/src/register/mvien.rs
@@ -1,0 +1,157 @@
+//! mvien register
+
+use crate::bits::{bf_extract, bf_insert};
+use riscv_pac::result::{Error, Result};
+use riscv_pac::InterruptNumber;
+
+#[cfg(target_arch = "riscv32")]
+const MASK: usize = 0xffff_e222;
+#[cfg(not(target_arch = "riscv32"))]
+const MASK: usize = 0xffff_ffff_ffff_e222;
+
+read_write_csr! {
+    /// `mvien` register
+    Mvien: 0x308,
+    mask: MASK,
+}
+
+read_write_csr_field! {
+    Mvien,
+    /// Alias of `mie.SSIE`
+    ssoft: 1,
+}
+
+read_write_csr_field! {
+    Mvien,
+    /// Alias of `mie.STIE`
+    stimer: 5,
+}
+
+read_write_csr_field! {
+    Mvien,
+    /// Alias of `mie.SEIE`
+    sext: 9,
+}
+
+impl Mvien {
+    /// Represents the minimum interrupt of the unlabelled virtual interrupt range.
+    pub const MIN_INTERRUPT: usize = 13;
+    /// Represents the maximum interrupt of the unlabelled virtual interrupt range.
+    #[cfg(target_arch = "riscv32")]
+    pub const MAX_INTERRUPT: usize = 31;
+    /// Represents the maximum interrupt of the unlabelled virtual interrupt range.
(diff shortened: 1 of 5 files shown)


romancardenas (changes requested):

riscv/src/register/mvienh.rs
```diff
@@ -0,0 +1,109 @@
...
+    /// Check if a specific core interrupt source is enabled.
+    #[inline]
+    pub fn is_enabled<I: InterruptNumber>(&self, interrupt: I) -> bool {
+        let n = interrupt.number();
+        Self::is_valid_interrupt(n) && bf_extract(self.bits, Self::shift_interrupt(n), 1) != 0
+    }
+
+    /// Enable a specific core interrupt source.
+    #[inline]
+    pub fn enable<I: InterruptNumber>(&mut self, interrupt: I) {
```
romancardenas: Same here

riscv/src/register/mvienh.rs
```diff
@@ -0,0 +1,109 @@
...
+        let n = interrupt.number();
+
+        if Self::is_valid_interrupt(n) {
+            self.bits = bf_insert(self.bits, Self::shift_interrupt(n), 1, 1);
+        }
+    }
+
+    /// Disable a specific core interrupt source.
+    #[inline]
+    pub fn disable<I: InterruptNumber>(&mut self, interrupt: I) {
```
romancardenas: Same here

romancardenas (commented):
What do you think about returning results to notify users about potential issues with their interrupt enum?

riscv/src/register/mvien.rs
```diff
@@ -0,0 +1,160 @@
...
+    #[inline]
+    pub const fn is_valid_interrupt(int: usize) -> bool {
+        matches!(int, 1 | 5 | 9 | Self::MIN_INTERRUPT..=Self::MAX_INTERRUPT)
+    }
+
+    /// Check if a specific core interrupt source is enabled.
+    ///
+    /// Returns `Error` if the interrupt number is invalid.
+    #[inline]
+    pub fn is_enabled<I: InterruptNumber>(&self, interrupt: I) -> Result<bool> {
```
romancardenas: I personally prefer this method like:

```suggestion
    pub fn is_enabled<I: InterruptNumber>(&self, interrupt: I) -> bool {
```

If the interrupt is invalid, it returns `false`, so I guess that is more than enough :)

riscv/src/register/mvienh.rs
```diff
@@ -0,0 +1,122 @@
...
+    #[inline]
+    pub const fn shift_interrupt(int: usize) -> usize {
+        int.saturating_sub(Self::INTERRUPT_SHIFT)
+    }
+
+    /// Check if a specific core interrupt source is enabled.
+    ///
+    /// Returns `Error` if the interrupt number is invalid.
+    #[inline]
+    pub fn is_enabled<I: InterruptNumber>(&self, interrupt: I) -> Result<bool> {
```
romancardenas: Same here. For this check function, I think we can skip the `Result` thing.

romancardenas (commented):
Let me know what you think. If you agree with my comments, apply the change and let's merge it!

0xllx0:
> Let me know what you think. If you agree with my comments, apply the change and let's merge it!

Awesome! I applied the changes. Let me squash down all of the fixup commits.

PR #368, opened 2025-11-08
Implement DPC CSR for RISC-V

KushalMeghani1644:
This PR implements the `DPC` CSR for RISC-V

Related to #1

### DIFF
--- a/riscv/src/register/dpc.rs
+++ b/riscv/src/register/dpc.rs
@@ -18,12 +18,14 @@ mod tests {
 
     #[test]
     fn test_dpc_bits_roundtrip() {
-        (0..=usize::BITS).map(|r| ((1u128 << r) - 1) as usize).for_each(|pc| {
-            // ensure lowest bit is cleared
-            let exp_pc = pc & !1usize;
-            let dpc = Dpc::from_bits(pc);
-            assert_eq!(dpc.bits(), exp_pc);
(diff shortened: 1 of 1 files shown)


KushalMeghani1644:
Great! @romancardenas  is there any improvements to be made? ~~or is just a quick merge?~~ :)

romancardenas:
Let's leave a few days in case someone from the WG wants to do a second review.

riscv/src/register/dpc.rs
```diff
@@ -0,0 +1,25 @@
+//! dpc register — Debug PC (0x7b1)
+
+read_write_csr! {
+    /// Debug PC Register
+    Dpc: 0x7b1,
+    mask: !1usize,
```
rmsyn: What are you basing this alignment on?

rmsyn (commented):
Small non-blocking suggestions, otherwise LGTM.

riscv/src/register/dpc.rs
KushalMeghani1644: The Debug Spec defines `dpc` as holding the address of the next instruction, and according to the base ISA alignment rules (for the C extension), instruction addresses are always 2-byte aligned.
So, bit 0 of `dpc` is always 0, which is why the mask `!1usize` is applied.  Hope this clears out what you were concerned about :)

KushalMeghani1644:
Thanks!

PR #370, opened 2025-11-13
Implement dscratch0 and dscratch1 CSRs for RISC-V

KushalMeghani1644:
This PR implements the dscratch0 and dscratch 1 CSRs

Related to #1

### DIFF
--- a/riscv/CHANGELOG.md
+++ b/riscv/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 - Add `dcsratch0` and `dscratch1` CSRs
+- Add new `read-write_csr_as_usize` macro for registers
 - Add `dpc` CSR support for RISC-V
 - Add Mtopi
 - Added DCSR (Debug Control and Status Register) CSR support for the RISC-V


romancardenas:
Maybe this is a good opportunity for a macro like `read_write_csr_as_usize` something like [this](https://github.com/rust-embedded/riscv/blob/78c9298ffa056d2c8404f9f6b0b00fbd33abd0f4/riscv/src/register/macros.rs#L113) but with writing. Don't you think?

Also, rebase this branch with master. It should pass all the CI checks.

KushalMeghani1644:
Thanks! and that's a good idea! I will soon push the changes with this macro added and being used for the dscratch0 and 1 codes! I'll also rebase with master so the CI is happy again :)

KushalMeghani1644:
Oops! Seems like I broke something... I'll fix this ASAP

riscv/src/register/dscratch0.rs
```diff
@@ -0,0 +1,17 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_dscratch0_read_write() {
+        for i in 0..usize::BITS {
+            let val = 1usize << i;
+            let _ = unsafe { try_write(val) };
```
romancardenas: Make the test check that the result is an Unimplemented error at least

romancardenas (changes requested):

KushalMeghani1644:
Thanks for the review! @romancardenas I have updated the tests as instructed!

romancardenas (approved):

PR #375, opened 2025-11-26
Use `riscv-types` instead of `riscv-pac`

romancardenas:
This PR deprecates `riscv-pac` and uses `riscv-types` instead. As a result, we need to bump the MSRV of several crates to 1.81


romancardenas:
@rmsyn  @MabezDev could you take a look to this PR? I would like to push a new release to crates.io and I think it makes sense to include this change.

MabezDev (approved):
LGTM, sorry for radio silence 😅

PR #380, opened 2025-12-18
Prepare for release

romancardenas:
This PR is ready for a new batch of releases in `crates.io`. @rmsyn @MabezDev could you check it out?

I am considering the MSRV bump to 1.81 as a breaking change, thus I am increasing the minor version of `riscv-peripheral` and `riscv-rt`. Let me know if you think we could leave it as a patch version.


MabezDev (approved):
LGTM, thanks!

PR #386, opened 2026-01-11
riscv-macros: Move `post_init` and `entry` attributes

romancardenas:
This PR is part of the rework for procedural macros of the RISC-V crates. Basically, it migrates `riscv_rt_macros::post_init` to `riscv_macros::post_init`.

The behavior of the macro remains mainly identical. However, I worked on a more modular implementation, with the idea of reusing as much code as possible once this PR is merged and we move `entry`, `exception`, etc. Hopefully, it will be easier in the future to maintain all the `riscv-rt` macros :)

I also modified the `multi-hart` QEMU example to illustrate how the macro works (and assert that it works properly).

Let me know what you think!


tests-qemu/examples/multi_hart.rs
```diff
@@ -104,21 +104,24 @@ _mp_hook:
 "#
 );
 
-#[entry]
-fn main(hartid: usize) -> ! {
```
rmsyn, line 108: Thanks for adding this test case to the QEMU test suite.

Maybe add cases for positive/negative results for each of the `check_*` functions to the test-trybuild suite?

romancardenas: Sure! There are already a few, but I am also thinking on a new error handling mechanism for these macros that stores multiple errors and throws them in batch. In this way, developers will able to see all the errors with just one attempt

romancardenas:
@rmsyn check it out, I think the new error handling is way better than in previous versions!

romancardenas:
(of course Rust 1.93 came with new clippy lints and compile error messages that broke our CI...)

romancardenas:
Now, `entry` is also in `riscv-macros`. I added `trybuild` tests to check that everything works as expected.

romancardenas:
Closed as per #389

PR #387, opened 2026-01-13
Make ARCH_WIDTH byte aligned instead of bit aligned

### LINKED ISSUE #385: riscv-rt: link.x uses ARCH_WIDTH bits as byte units
riscv-rt's build.rs replaces ${ARCH_WIDTH} with the architecture's 32/64/128 bit size. The link.x.in script treats this value as bytes rather than bits. For example:
```
/* ${ARCH_WIDTH}-byte align the end (VMA) of this section.
   This is required by LLD to ensure the LMA of the following .data
   section will have the correct alignment. */
. = ALIGN(${ARCH_WIDTH});
```
This extra alignment is unnecessary space. Eliminating it would save space.

kurtjd:
#385 pointed out the build script is passing `ARCH_WIDTH` in bits rather than bytes, so this PR simply divides width by 8 to correctly pass width in bytes. Does indeed seem to help reduce binary size slightly (the one example I tested it on reduced `.rodata` by 4 bytes).

Additionally bumps version to `0.17.1`.

Resolves #385

### DIFF
--- a/riscv-rt/CHANGELOG.md
+++ b/riscv-rt/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## v0.17.1 - 2026-01-13
+
+### Fixed
+
+- Fix `build.rs` so `ARCH_WIDTH` is passed to linker script in size of bytes (instead of bits)
+
 ## v0.17.0 - 2025-12-19
 
 ### Added
--- a/riscv-rt/Cargo.toml
+++ b/riscv-rt/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "riscv-rt"
-version = "0.17.0"
+version = "0.17.1"
 rust-version = "1.81"
 repository = "https://github.com/rust-embedded/riscv"
 authors = ["The RISC-V Team <risc-v@teams.rust-embedded.org>"]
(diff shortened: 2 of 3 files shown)


riscv-rt/CHANGELOG.md
romancardenas, line 9:
```suggestion

## v0.17.1 - 2026-01-13

### Fixed

- Fix `build.rs` so `ARCH_WIDTH` is passed to linker script in size of bytes (instead of bits)
```

riscv-rt/CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- Fix `build.rs` so `ARCH_WIDTH` is passed to linker script in size of bytes (instead of bits)
```
romancardenas:
```suggestion
```

romancardenas (changes requested):
Thanks!

Please, address the comments. Also. if you modify `riscv-rt/Cargo.toml` to point to version `0.17.1`, I will publish a patch today

riscv-rt/CHANGELOG.md
kurtjd: Sorry about that, thanks for catching it.

riscv-rt/CHANGELOG.md
kurtjd: Done. Bumped to `v0.17.1` in Cargo.toml as well.

romancardenas (approved):

PR #388, opened 2026-01-16
`riscv-macros`: Move `entry` attribute macro

romancardenas:
This PR is a continuation of #386 . It moves the `entry` macro from `riscv-rt-macros` to `riscv-macros`.

It also serves as an example of how the new macro refactoring allows us to reuse the code for all the attributes, easing the maintenance efforts.


MabezDev (approved):
LGTM, thanks!

romancardenas:
@MabezDev can you check #386 and let me know if the new error handling is compatible with `esp-riscv-rt`? It should be, but just in case.

romancardenas:
closing in favor of #386

PR #389, opened 2026-02-02
`riscv-macros`: add all the macros required by `riscv-rt`

romancardenas:
This PR reimplements all the macros previously on `riscv-rt-macros` to `riscv-macros`.

The macro code is now more modular, which will potentially help us maintaining the source code as well as adding new attribute macros if needed. Note also that, now, the attribute macros for `riscv-rt` functions output multiple errors in a single run. This will improve the developer experience, as with a single compilation attempt, they can see all the errors and address them before re-trying.

This PR also deprecates `riscv-rt-macros` in favor of `riscv-macros`.

Now, `riscv-rt` does not re-export the `pre_init` attribute macro, which was already deprecated. Users now have to use `core::arch::global_asm!`

This PR substitutes #386 , as it contains the same changes but with a clean commit history. I hope it is relatively easy to review now!


MabezDev (approved):
Macros are not my area of expertise, but this looks okay from my first glance, and reasonably well tested. Thanks!

Issue #155, opened 2022-01-20
Prune unused symbols


piegamesde:
I'm trying to optimize my binary size. For this, I override all default functions that I don't need with stubs. However, most of these symbols are still present in my binary. Is there a way to get rid of them?

As far as I can tell, using weak symbols could resolve this, but I don't know much about linking. (Incomplete) list of symbols that I'd like to get rid of:

- `__wfi`
- `__read_mhartid`
- `__read_mcause`
- `default_mp_hook`
- `DefaultExceptionHandler`
- `DefaultInterruptHandler`
- `default_setup_interrupts`
- `default_start_trap`

PR #390, opened 2026-02-09
`riscv-rt`: new attribute macro to define custom `_setup_interrupts` function.

romancardenas:
This PR provides the following changes:

1. New `custom-setup-interrupts` feature to opt-out the default implementation of `_setup_interrupts`
2. Now, instead of defining a `_default_setup_interrupts` symbol that acts as a weak implementation of `_default_setup_interrupts`, `riscv-rt` now directly defines `_setup_interrupts` (protected under a feature gate). This reduces the amount of code in the final binary when users define their custom interrupt setup routine.
3. Now, `_setup_interrupts` can accept an `usize` input argument that corresponds to the ID of the running hart. This might be useful in multi-hart targets to define hart-specific interrupt setup routines.
4. New `riscv_macros::setup_interrupts` attribute macro for aiding users defining their custom interrupt setup routine.

## About feature naming

I think `custom-setup-interrupts` is a clear name for the feature, as it should be activated whenever users want to implement their custom interrupt setup routine. However, previous similar features follow a `no-*` name (e.g., `no-interrupts` or `no-exceptions`). Let me know if you prefer to stick to the `no-*` naming or if you prefer using `custom-*`.

While I am currently more inclined to `custom-*`, I would like to know you opinion.

### DIFF
--- a/riscv-macros/CHANGELOG.md
+++ b/riscv-macros/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- New `setup_interrupts` macro for custom interrupt setup routines.
 - New `exception`, `core_interrupt`, and `external_interrupt` macros for trap handlers.
 - New `entry` macro for the Rust entry point (for `riscv-rt`).
 - New `rvrt-u-boot` feature to adapt `entry` macro for U-Boot.
--- a/riscv-macros/Cargo.toml
+++ b/riscv-macros/Cargo.toml
@@ -12,6 +12,9 @@ repository = "https://github.com/rust-embedded/riscv"
 version = "0.4.1"
 edition = "2021"
 
+[package.metadata.docs.rs]
+features = ["riscv-rt"]
+
 [lib]
 proc-macro = true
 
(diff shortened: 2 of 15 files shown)


romancardenas:
@MabezDev what do you think about calling the feature `custom-setup-interrupts` instead of `no-setup-interrupts`? It breaks the current naming convention, but I think it is clearer the purpose of this feature.

thejpster:
Wouldn’t the default function be garbage collected by the linker when it observes that it’s unreachable when it’s replaced by another function?

If not, I have problems over in aarch32 land…

romancardenas:
This full story started with #155 

It looks like "default" symbols are not optimized out, and when you override them, you end up having both routines. We have been working on reducing/reusing default symbols to minimize this binary size overhead since then.

As a side note, we experimented with weak symbols in assembly (that was our first approach). However, using weak symbols currently breaks LTO (#247)

romancardenas:
@MabezDev @rmsyn I'd appreciate a review so this can be merged and I can move to other contributions

MabezDev (approved):
> what do you think about calling the feature custom-setup-interrupts

I agree, this is better.

`procmacros` aren't my expertise, but this looks solid. Thanks!

PR #398, opened 2026-04-29
`riscv-rt`: use `pac_enum` macro for exception handling

romancardenas:
Minor PR. Instead of defining all the required boilerplate code for handling exceptions, we let the `riscv::pac_enum` macro do the work for us.

It also fixes the versioning of `riscv-rt`, getting prepared for a new release.


romancardenas:
@MabezDev can you take a quick look so I can ship a new release by the end of this week?

MabezDev (approved):
At RustWeek currently, so this a is bit of blind +1, but I don't want to block your release schedule.

PR #403, opened 2026-06-16
Systematically define bit mask/offset/size for fields in CSRs

jsgf:
This PR systematically adds associated constants for each bitfield in the CSRs. Most of the heavy
lifting was done by extending the existing bitfield definition macros (read_only_csr_field! /
read_write_csr_field!), but there were also specific additions for:
- mstatus/sstatus:  UXL, SXL, SBE, MBE (RV64 fields implemented outside the macros)
- mtvec/stvec:  BASE_SHIFT/BASE_MASK (the MODE field is already macro-generated via TRAP_MODE_*)
- pmpcfg:  R/W/X/PERMISSION/A/L byte-layout constants on the Pmp type (PMP uses enums, not the field
   macros)

These are just new definitions, so no changes to existing APIs. It's primarily to make these constants available to inline asm.


romancardenas (approved):
Looks good to me! Thank you very much

PR #404, opened 2026-07-20
Add mireg CSR (#226)

reubenueben:
Adds the `mireg` (CSR 0x351) register

Related issues : #226  #1

### DIFF
--- a/riscv/src/register/mireg.rs
+++ b/riscv/src/register/mireg.rs
@@ -1,0 +1,14 @@
+//! 'mireg' register (Machine Indirect Register)
+//!
+//! CSR address: 0x351
+//!
+//! The `mireg` register provides access to an indirect register
+//! selected by `miselect`.
+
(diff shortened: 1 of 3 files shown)


riscv/src/register/mireg.rs
```diff
@@ -0,0 +1,14 @@
+//! 'mireg' register (Machine Interrupt Register)
```
romancardenas: The CSR is called "Machine Indirect Register": https://docs.riscv.org/reference/isa/v20260120/priv/priv-csrs.html#:~:text=second%20trap%20value.-,Machine,-Indirect

romancardenas (changes requested):

romancardenas:
Please fix formatting. Then, squash all your commits so we can keep a clean commits history

romancardenas (approved):

PR #410, opened 2026-08-06
riscv: Add `From<usize>` and `From<CSR>` on `csr!` types

JurajSadel:
Every type generated by the `csr!` macro already exposes `from_bits()` and `bits()`. This adds the corresponding standard conversions.


riscv/src/register/macros.rs
```diff
@@ -621,6 +621,20 @@ macro_rules! csr {
                 Self::BITMASK
             }
         }
+
+        impl From<usize> for $ty {
+            #[inline]
+            fn from(bits: usize) -> Self {
+                Self::from_bits(bits)
+            }
+        }
```
romancardenas, line 630: I'm not sure about this. Maybe it is better to implement TryFrom, where both Ok and Err variants are Self. However, When Err is returned, it means that the provided binary value has been modified by the register mask (i.e., it is not exactly the same as the provided input)

Thoughts?

romancardenas (changes requested):

riscv/src/register/macros.rs
JurajSadel: Agreed. My use case doesn't really need this after all - I will close this PR. Thanks for the review.

PR #414, opened 2026-08-07
riscv: add misa write side

JurajSadel:


riscv/src/register/misa.rs
```diff
@@ -16,6 +16,11 @@ read_only_csr! {
     sentinel: 0,
 }
 
+// `misa` is WARL: a hart may ignore the write entirely, and disabling an
+// extension while its state is live is up to the caller. Reading the register
+// back is the only way to learn what took effect.
+write_csr_as!(Misa, 0x301);
```
romancardenas: Instead of this, would it be possible to use the `read_write_csr!` macro?

romancardenas (changes requested):

riscv/src/register/misa.rs
JurajSadel: Yes, works for my use case, thank you!

riscv/src/register/misa.rs
```diff
@@ -1,19 +1,17 @@
 //! misa register
 
 #[cfg(target_arch = "riscv32")]
-read_only_csr! {
+read_write_csr! {
     /// `misa` register
     Misa: 0x301,
     mask: 0xc3ff_ffff,
-    sentinel: 0,
```
romancardenas, line 8: I understand the currency versión of the read_write macro does not accept sentinel values right?

Could you modify the macro so it implements the sentinel functionality?

romancardenas (changes requested):

riscv/src/register/misa.rs
JurajSadel: Sorry for late response, I was on vacation and this PR felt out of my radar.

romancardenas (approved):

